PR #62, opened 2025-12-25
Add apk package manager support

### LINKED ISSUE #61: Support OpenWRT 25.12
##### SUMMARY
The first release candidate of OpenWRT 25.12 was just released. For me it looks like this is the only significant change for us:
-  Switch package manager from opkg to apk

https://openwrt.org/releases/25.12/notes-25.12.0-rc1#highlights_in_openwrt_2512

I don't know anything about it, yet, but to keep existing playbooks compatible I'd favor to add an alias "apk" for opkg (or the other way round, as apk is the future) and add support for apk to this module (with some autodetection which one to use or so).

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
opkg

kbialek:
##### SUMMARY
Add `apk` package manager support, which is used by default in openwrt 25.12

Fixes #61 

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME
apk

##### ADDITIONAL INFORMATION
1. `init` role is updated to prefer `apk` over `opkg` when installing recommended packages. 
    1. `apk` keeps its cache up to date, hence no need to check if the cache refresh is needed
2. added new molecule platform `openwrt_x86_64-25.12.0-rc1`
3. added new tests for `apk` module verification
4. updated collection is working fine on my test router with openwrt 25.12 installed, as well as on my "prod" routers running openwrt 24.10

### DIFF
--- a/plugins/modules/apk.py
+++ b/plugins/modules/apk.py
@@ -79,20 +79,4 @@ EXAMPLES = r"""
     no_cache: true
 """
 
-RETURN = r"""
-stdout:
-  description: apk standard output.
-  returned: always
-  type: str
-  sample: Foo foo foo
-stderr:
-  description: apk standard error.
-  returned: always
-  type: str
-  sample: ""
-rc:
-  description: apk return code.
-  returned: always
-  type: int
-  sample: 0
-"""
+RETURN = r""""""


plugins/modules/apk.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/sh
+# shellcheck shell=ash
+# Copyright (c) 2024 Krzysztof Bialek/Markus Weippert
```
russoz:
```suggestion
# Copyright (c) 2025 Krzysztof Bialek/Markus Weippert
```

plugins/modules/apk.sh
```diff
@@ -0,0 +1,96 @@
...
+        if ! query_package "$pkg"; then
+            pkgs_to_install="$pkgs_to_install $pkg"
+        fi
+    done
+
+    [ -n "$pkgs_to_install" ] || return 0
+
+    [ -n "$_ansible_check_mode" ] || {
+        # shellcheck disable=SC2086
+        try apk add $update_cache $no_cache $force_broken_world $pkgs_to_install
```
russoz: Probably only `pkgs_to_install` needs the SC disable. In that case, we should apply the good practice to the other params.
```suggestion
        try apk add "$update_cache" "$no_cache" "$force_broken_world" $pkgs_to_install
```

plugins/modules/apk.py
```diff
@@ -0,0 +1,81 @@
...
+# Copyright (c) 2024 Krzysztof Bialek/Markus Weippert
+# GNU General Public License v3.0 (see https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import annotations
+
+DOCUMENTATION = r"""
+module: apk
+short_description: Manage packages with apk on OpenWrt
+description:
+  - The M(community.openwrt.apk) module manages packages on OpenWrt using the apk package manager (available in OpenWrt 24.10+).
```
russoz: I believe this should be:
```suggestion
  - The M(community.openwrt.apk) module manages packages on OpenWrt using the apk package manager (available in OpenWrt 25.12+).
```

plugins/modules/apk.sh
```diff
@@ -0,0 +1,96 @@
...
+
+    # Map boolean update_cache to apk flag
+    if [ "$update_cache" = "True" ] || [ "$update_cache" = "yes" ]; then
+        update_cache="--update-cache"
+    else
+        update_cache=""
+    fi
+
+    # Map no_cache (common in apk)
+    if [ "$no_cache" = "True" ] || [ "$no_cache" = "yes" ]; then
```
russoz: I _believe_ the `wrapper.sh` will normalize bool params, btu not sure exactly what to. Give it a read. As soon as we have tests ready this will be easier to assess.

extensions/molecule/default/verify.yml
```diff
@@ -131,3 +131,30 @@
       ansible.builtin.file:
         path: /tmp/openwrt_test_file.txt
         state: absent
+
+    - name: Apk install package
+      community.openwrt.apk:
+        name: htop
+      when: _apk_available.rc == 0
```
Jonny007-MKD: IMO it would be nicer to use a `block` to only check for `_apk_available` once

plugins/modules/apk.py
```diff
@@ -0,0 +1,82 @@
...
+    update_cache: true
+
+- name: Install package without using local cache
+  community.openwrt.apk:
+    name: tcpdump
+    state: present
+    no_cache: true
+"""
+
+RETURN = r""""""
```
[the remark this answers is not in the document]
russoz: Errr... I correct myself: the pylint sanity check does enforce that (on .py files, at least)

extensions/molecule/default/molecule.yml
```diff
@@ -2,6 +2,10 @@
 driver:
   name: docker
 platforms:
+  - name: openwrt_x86_64-25.12.0-rc1
+    image: openwrt/rootfs:x86_64-25.12.0-rc1
+    command: sh -c "mkdir -p /var/lock; sleep infinity"
```
russoz: Have you tried without the `/var/lock`? Does `apk` require it as well?

plugins/modules/apk.py
```diff
@@ -0,0 +1,82 @@
...
+from __future__ import annotations
+
+DOCUMENTATION = r"""
+module: apk
+short_description: Manage packages with apk on OpenWrt
+description:
+  - The M(community.openwrt.apk) module manages packages on OpenWrt using the apk package manager (available in OpenWrt 25.12+).
+  - It can install and remove packages.
+author: Krzysztof Bialek (@kbialek)
+version_added: 0.3.0
```
russoz: Sorry, I planned that poorly: this now needs to be updated to 0.4.0, as 0.3.0 has already been released.

plugins/modules/apk.sh
```diff
@@ -0,0 +1,96 @@
...
+"
+
+query_package() {
+    # apk info returns 0 if installed, 1 if not
+    apk info -e "$1" >/dev/null 2>&1
+}
+
+install_packages() {
+    local _IFS pkg pkgs_to_install
+    _IFS="$IFS"; IFS=","; set -- $name; IFS="$_IFS"
```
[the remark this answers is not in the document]
russoz: And, to @kbialek question: as you can notice from the previous comment, my understanding is that it will be a string with space-delimited items, but this should really be tested.

roles/init/tasks/packages-apk.yml
```diff
@@ -0,0 +1,17 @@
...
+- name: Check for openssl
+  community.openwrt.command:
+    cmd: which openssl
+  register: _openssl
+  check_mode: false
+  changed_when: false
+  failed_when: false
+
+- name: Install recommended packages if openssl is not available
+  community.openwrt.apk:
```
russoz: It looks like there has been no cache update for `apk` as it does with `opkg`. With `opkg` that would fail. Could you please confirm it works for `apk`?

russoz (commented):
Hi @kbialek 

Second round of comments.

For the `init` role, I was thinking whether it would be better to have a variable that would contain either `opkg` or `apk` and then:
```
- include_tasks: packages-{{ pkg_type }}
```
(just an example, I would name the var differently)

Anyways, that's just an idea.

extensions/molecule/default/verify.yml
kbialek: Good point!

plugins/modules/apk.sh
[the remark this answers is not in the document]
kbialek: @russoz I tested it after changing the type in apk.sh like this `name=pkg/list/r`. The result are as follows:
1. When a single string element is set, e.g.
```
  community.openwrt.apk:
    name: coreutils-base64,coreutils-md5sum,coreutils-sha1sum
```
then
`name` is set to `coreutils-base64,coreutils-md5sum,coreutils-sha1sum` in the input JSON, and is passed through as a plain string

2. When a list of string is set, e.g.
```
          community.openwrt.apk:
            name:
              - curl
              - wget
```
then
`name` is set to `["curl", "wget"]` in the input JSON, but the wrapper fails to process it.
The `name` shell var is being set to `J_A1`, which is jshn internal array pointer I belive.
The following change in the `wrapper.sh` fixes it (fix suggested by gemini)
```
        case "$_is_type" in
            array)
                # 1. Select the array to move the pointer inside it
                json_select "$_alias"
                
                # 2. Iterate through elements (jshn uses 1-based indexing for arrays)
                local i=1
                local entry
                _value=""
                while json_get_var entry "$((i++))"; do
                    _value="${_value:+$_value }$entry"
                done
                
                # 3. Move back up to the parent level
                json_select ..

                # 4. Export the space-separated string
                eval "export -- $_var=\"\$_value\" _type_$_var=\"array\""
                ;;
            object)
                [ "$_type" = "$_is_type" -o "$_type" = "any" ] ||
                    fail "$_var must be $_type"
                eval "export -- _type_${_var}=\"\$_is_type\""
                json_add_string "_$_var" "$_alias";;
```
This sets the name shell var to `curl wget` string

What do you think about this change?

roles/init/tasks/packages-apk.yml
kbialek: Yes, `apk` takes care of its initial cache update. It can be easily tested using the following commands
```
rm -rf /var/cache/apk/
apk -vv add bc
```

extensions/molecule/default/molecule.yml
```diff
@@ -2,6 +2,10 @@
 driver:
   name: docker
 platforms:
+  - name: openwrt_x86_64-25.12.0-rc1
+    image: openwrt/rootfs:x86_64-25.12.0-rc1
+    command: sh -c "sleep infinity"
```
russoz: Since it is only one simple command, you can run it directly. ;-)
```suggestion
    command: sleep infinity
```

plugins/modules/apk.sh
```diff
@@ -0,0 +1,75 @@
...
+# Copyright (c) 2025 Krzysztof Bialek/Markus Weippert
+# GNU General Public License v3.0 (see https://www.gnu.org/licenses/gpl-3.0.txt)
+
+PARAMS="
+    name=pkg/str/r
+    state/str//present
+    update_cache/bool
+    no_cache/bool
+    force_broken_world/bool
+"
```
russoz: As soon as #67 is merged, this will need to go into a `init()` function.

roles/init/tasks/main.yml
```diff
@@ -18,11 +18,23 @@
...
+    cmd: which apk
+  register: _apk_available
+  check_mode: false
+  changed_when: false
+  failed_when: false
+  when: openwrt_install_recommended_packages | bool
+
+- name: Check for package manager availability
+  ansible.builtin.set_fact:
+    _package_manager_name: "{{ 'apk' if _apk_available.rc == 0 else 'opkg' if _opkg_available.rc == 0 else None }}"
```
russoz: As indicated previously, may I suggest this becoming an "official" fact coming from the role? Name should be `openwrt_`-something. Maybe `openwrt_package_manager`?

Then you use it in the `verify.yml` test.

plugins/modules/apk.sh
```diff
@@ -0,0 +1,75 @@
...
+}
+
+main() {
+    case "$state" in
+        present|installed|absent|removed) :;;
+        *) fail "state must be present or absent";;
+    esac
+
+    [ "$update_cache" = "1" ] && update_cache="--update-cache" || update_cache=""
+    [ "$no_cache" = "1" ] && no_cache="--no-cache" || no_cache=""
```
russoz: I think these two options should be mutually exclusive, no?

plugins/modules/apk.sh
```diff
@@ -0,0 +1,75 @@
...
+        done
+    }
+    changed
+}
+
+main() {
+    case "$state" in
+        present|installed|absent|removed) :;;
+        *) fail "state must be present or absent";;
+    esac
```
russoz: Like a previous comment, after #67 is merged this should go into `validate() {}`

russoz (commented):
Hi @kbialek 

Thanks for your continued work on this! I got another round of comments but it is really looking good now. We are very close to merging this PR.

plugins/modules/apk.py
```diff
@@ -0,0 +1,82 @@
...
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+options:
+  name:
+    description:
+      - Name of the package(s) to install or remove.
+      - Multiple packages can be specified as a comma-separated list.
```
russoz: Are we handling spaces around the comma? If not, let the user know.

plugins/modules/apk.sh
russoz: Just merged it!

plugins/modules/apk.py
```diff
@@ -0,0 +1,82 @@
...
+    update_cache: true
+
+- name: Install package without using local cache
+  community.openwrt.apk:
+    name: tcpdump
+    state: present
+    no_cache: true
+"""
+
+RETURN = r""""""
```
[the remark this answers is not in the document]
kbialek: I've just added stdout, stderr and rc return vars

extensions/molecule/default/verify.yml
Jonny007-MKD: tests for `check_mode` are missing. I see 4 cases:
| pre-condition | action with check_mode | yielded result | verification |
| ---           | ---           |---           | ---           |
| before installing htop | install htop | a modification | htop is not installed afterwards |
| after installing htop |install htop | no modification| htop is still installed |
| after installing htop |uninstall htop | a modification | htop is still installed |
| after uninstall htop | uninstall htop | no modification | htop is still not installed :) |

extensions/molecule/default/verify.yml
```diff
@@ -131,3 +131,70 @@
...
+
+        - name: Apk assert install package stderr
+          ansible.builtin.assert:
+            that: "not _apk_result.stderr"
+            fail_msg: "Unexpected stderr {{ _apk_result.stderr }}"
+
+        - name: Apk check package is installed
+          community.openwrt.command:
+            cmd: which htop
+            uses_shell: true
```
russoz: We don't need shell to run that, but TBH not sure that makes much of a difference in our context. In any case, just to be consistent with what one would do at `ansible.builtin.command`:
```suggestion
            cmd: which htop
```

plugins/modules/apk.sh
```diff
@@ -0,0 +1,102 @@
...
+    RESPONSE_VARS="
+        stdout/str/a
+        stderr/str/a
+        rc/int/a
+    "
+
+    export stdout=""
+    export stderr=""
+    export rc="0"
+    out="$(mktemp)" && err="$(mktemp)"
```
russoz: There is no dependency between the two actions:
```suggestion
    out="$(mktemp)"
    err="$(mktemp)"
```

plugins/modules/apk.py
```diff
@@ -0,0 +1,98 @@
...
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+options:
+  name:
+    description:
+      - Name of the package(s) to install or remove.
+      - Multiple packages can be specified as a comma-separated list with no spaces, e.g. curl,wget
```
russoz: (my two canned responses from reviewing PRs in community.general ;-) )

Please refer to the [linguistic guidelines](https://docs.ansible.com/projects/ansible/latest/dev_guide/style_guide/index.html#linguistic-guidelines) for the documentation, as well as the guide to [preferred spelling](https://docs.ansible.com/projects/ansible/latest/dev_guide/style_guide/spelling_word_choice.html).

Please use semantic markup as documented in:
https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#semantic-markup-within-module-documentation

```suggestion
      - Multiple packages can be specified as a comma-separated list with no spaces, for example V(curl,wget).
```

plugins/modules/apk.py
```diff
@@ -0,0 +1,98 @@
...
+      - removed
+    default: present
+  update_cache:
+    description:
+      - Update the package cache (C(apk update)) before performing the operation. Mutually exclusive with (C(no_cache)).
+    type: bool
+    default: false
+  no_cache:
+    description:
+      - Do not use a local cache files, fetch from index directly. Mutually exclusive with (C(update_cache)).
```
russoz:
```suggestion
      - Do not use a local cache files, fetch from index directly. Mutually exclusive with (O(update_cache)).
```

plugins/modules/apk.py
```diff
@@ -0,0 +1,98 @@
...
+
+RETURN = r"""
+stdout:
+  description: apk standard output.
+  returned: always
+  type: str
+  sample: Foo foo foo
+stderr:
+  description: apk standard error.
+  returned: always
```
russoz: Probably:
```suggestion
  returned: on error
```
?

tests/sanity/ignore-2.17.txt
```diff
@@ -26,6 +28,8 @@ plugins/modules/uci.sh shebang
 plugins/modules/uci.sh validate-modules:invalid-extension
 plugins/modules/wrapper.sh shebang
 plugins/modules/wrapper.sh validate-modules:invalid-extension
+plugins/modules/apk.sh shellcheck:SC2086   # Double quote to prevent globbing and word splitting - https://www.shellcheck.net/wiki/SC2086
```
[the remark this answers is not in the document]
russoz: Sorry, I did not see this before start reviewing - I have answered that already.

extensions/molecule/default/verify.yml
kbialek: Removed

plugins/modules/apk.py
kbialek: I disagree here. It's possible that `apk` writes to `stderr` (maybe some warnings?) even though the overall command exit status is successful. Hence, `stderr` is always returned, likely it's empty, but is returned.

plugins/modules/apk.sh
```diff
@@ -0,0 +1,102 @@
...
+
+query_package() {
+    # apk info returns 0 if installed, 1 if not
+    apk info -e "$1" >/dev/null 2>&1
+}
+
+install_packages() {
+    local _IFS pkg pkgs_to_install
+    # shellcheck disable=SC2154,SC2086
+    _IFS="$IFS"; IFS=","; set -- $name; IFS="$_IFS"
```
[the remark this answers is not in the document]
russoz: Err, I beg to differ. It did work for me locally - but you MUST add that `;` after the last command. Maybe that was the issue?

plugins/modules/apk.py
russoz: Err, I did not pay attention enough earlier today. Actually `stdout`, `stderr`, `rc` and a number of other Return Values are deemed "common return values" and therefore are usually not documented on a per-module basis. 

https://docs.ansible.com/projects/ansible/latest/reference_appendices/common_return_values.html

In community.general we usually do not add those RVs to the RETURN documentation block - there might be some exceptions, of course. For stdout and stderr they are all going to say something like "the stdout/stderr of the command X or Y". So, this is cutting on the repetition a bit.

Sorry for not spotting this sooner.

kbialek: So you say I should remove them from the docu?

I took `apt` docs as inspiration https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/apt_module.html#return-values and they are explicitly listed there.

russoz:
@kbialek friendly check: are you holding back a couple of those changes, like the other day? Some of the comments marked 'resolved' still look the same. Not a problem, just wondering.

extensions/molecule/default/verify.yml
russoz: Still there.

plugins/modules/apk.py
russoz: Also this one is still there.

kbialek:
> @kbialek friendly check: are you holding back a couple of those changes, like the other day? Some of the comments marked 'resolved' still look the same. Not a problem, just wondering.

No, I pushed everything

plugins/modules/apk.py
russoz: One thing though: if you want to keep them, please add the `C( )` markup around the `apk` command, in all RVs.

roles/init/tasks/main.yml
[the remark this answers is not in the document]
russoz: I have not enforced changelog fragments yet because we have not released version 1.0.0, and we have had changes being added and later removed or further changed way too quickly.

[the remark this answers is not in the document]
russoz: A.k.a. there is no real promise of backward compatibility until 1.0.0

russoz (approved):

plugins/modules/apk.py
russoz: @Jonny007-MKD Assuming this comment is done. Feel free to "unresolve" if you want to add more to it.

roles/init/tasks/main.yml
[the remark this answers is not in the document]
russoz: I totally agree with that, but I will not enforce it until 1.0.0.

russoz:
Thanks @kbialek for your contribution!
Thanks @Jonny007-MKD @felixfontein for reviewing!

PR #79, opened 2026-01-04
[WIP] command: alternate way of not supporting check mode

russoz:
##### SUMMARY

As I did #78 it occurred to me that we might perform the check mode support enforcing more explicitly - instead of setting a variable, the module actually calls a function.

Not much change for the module:
- replace the variable with the function call, in `validate()` (after parameters are parsed)

Not much change for the wrapper:
- one variable less to keep track of
- the function testing check mode gets one notch simper (by not checking the var)

If we decide to adopt this, then the PR needs to be expanded to:
- wrapper:
  - remove `_support_check_mode()`
  - remove call to `_support_check_mode` (by the end of the wrapper)
- modules: ensure modules are using `no_check_mode` instead of setting `SUPPORTS_CHECK_MODE`

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

wrapper
command

### DIFF
--- a/plugins/modules/command.sh
+++ b/plugins/modules/command.sh
@@ -4,7 +4,6 @@
 # GNU General Public License v3.0 (see https://www.gnu.org/licenses/gpl-3.0.txt)
 
 init() {
-    SUPPORTS_CHECK_MODE=""  # effectively False
     PARAMS="
         cmd=raw_params=_raw_params/str/r
         uses_shell=_uses_shell/bool//false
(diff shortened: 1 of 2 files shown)


russoz:
> If I am not mistaken, python ansible modules also export `supports_check_mode=True` and ansible automatically skips them if it is false. Thus, the current approach seems closer to the original.

Well, that is true, but only inside the context of an `AnsibleModule()` call.

I got mixed feelings myself over this. Will let it open for another couple of days, if I cannot find a compelling reason to go forward with this, then I will cancel this PR.

PR #95, opened 2026-01-14
fix(copy): respect force=false when destination exists

### LINKED ISSUE #94: Bug: community.openwrt.copy module fails when force=false and the destination file already exists
##### SUMMARY
The community.openwrt.copy module fails when force: false is set and the destination file already exists. According to the documentation, the module should no-op in this situation, but instead returns an error.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.openwrt.copy

##### ANSIBLE VERSION

```
ansible [core 2.15.13]
  config file = /home/drwicid/code/drwicid/infra-ops/configuration/ansible.cfg
  configured module search path = ['/home/drwicid/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/drwicid/.local/lib/python3.9/site-packages/ansible
  ansible collection location = /home/drwicid/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/drwicid/.local/bin/ansible
  python version = 3.9.2 (default, Mar 20 2025, 02:07:39) [GCC 10.2.1 20210110] (/usr/bin/python3)
  jinja version = 3.1.6
  libyaml = True
```

##### COLLECTION VERSION

```
# /home/drwicid/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.openwrt 0.4.0  
```

##### CONFIGURATION

```paste below
CONFIG_FILE() = /home/drwicid/code/drwicid/infra-ops/configuration/ansible.cfg
EDITOR(env: EDITOR) = vim
HOST_KEY_CHECKING(/home/drwicid/code/drwicid/infra-ops/configuration/ansible.cfg) = False
```

##### OS / ENVIRONMENT

OpenWrt 24.10.2 on x86/64

##### STEPS TO REPRODUCE

Create file on target system in destination location, run playbook with task:

```yaml
name: Copy file with force disabled
community.openwrt.copy:
  content: ''
  dest: /etc/example.conf
  force: false
```

##### EXPECTED RESULTS

When force is false and the destination file exists:

- The module performs no action
- The task does not fail
- The task reports no change

Documentation states:
"If false, the file is only transferred if the destination does not exist."

##### ACTUAL RESULTS

Fails with error "file already exists"

```paste below
fatal: [router2]: FAILED! => {"changed": false, "dest": "/etc/example.conf", "md5sum": "d41d8cd98f00b204e9800998ecf8427e", "msg": "file already exists", "src": "/tmp/ansible-tmp-1768413735.8733706-783355-267621589263465/source"}
```

drwicid:
##### SUMMARY

Updates `copy.sh` to return 0 when the dest file exists, and force is not set to true.

Fixes #94

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

community.openwrt.copy

##### ADDITIONAL INFORMATION

After the change, when copy encounters an existing file and force is false/unset, return 0 results in no changes. The file is not replaced and ansible shows the task as ok/unchanged.

### DIFF
--- a/tests/integration/targets/copy/tasks/main.yml
+++ b/tests/integration/targets/copy/tasks/main.yml
@@ -84,6 +84,7 @@
 - name: Print content of remote file
   community.openwrt.command:
     cmd: cat /tmp/test_copy_remote_file.txt
+    uses_shell: true
   register: cat_remote
 
 - name: Assert content of copied file


plugins/modules/wrapper.sh
```diff
@@ -210,6 +210,15 @@ changed() {
     CHANGED="1"
 }
 
+unchanged() {
+    CHANGED=""
+}
+
+succeed() {
+    [ -n "$@" ] && MESSAGE="$@"
```
russoz: Err... nope. I was wrong (again). Sorry.

This should use `$*` instead of `$@` because the context is not of a list. We _do want_ the list of args to be interpolated into a string. Sorry, please rollback to how it was before.

tests/integration/targets/copy/tasks/main.yml
```diff
@@ -22,18 +22,18 @@
     that:
       - copy_out2 is not changed
 
-- name: "Copy different content to the same file again with force=false, should fail"
+- name: "Copy different content to the same file again with force=false, should not change"
   community.openwrt.copy:
     content: "different test content for copy_out3a"
     dest: "/tmp/test_copy_file.txt"
     force: false
   register: copy_out3a
```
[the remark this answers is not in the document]
russoz: No worries, all good. For me it's morning yet and I am probably one or two coffees short. ;-)

tests/integration/targets/copy/tasks/main.yml
```diff
@@ -22,18 +22,18 @@
...
-  ignore_errors: true
 
-- name: Assert copy_out3a failed
+- name: Assert copy_out3a was not changed
   ansible.builtin.assert:
     that:
-      - copy_out3a is failed
+      - copy_out3a is not changed
+      - copy_out3a is succeeded
```
russoz: This assertion is not wrong but it is redundant - without `ignore_errors=true` (removed line 31), if the copy fails, then the task fails, then the test fails.

tests/integration/targets/copy/tasks/main.yml
```diff
@@ -54,13 +54,14 @@
...
   register: copy_out3c
-  ignore_errors: true # @TODO bug reported in #94
 
-- name: Assert copy_out3c was changed
+- name: Assert copy_out3c was not changed
   ansible.builtin.assert:
     that:
       - copy_out3c is not changed
-      - copy_out3c is failed # @TODO bug reported in #94, it should not be failed
+      - copy_out3c is succeeded
```
russoz: same here - this is redundant

drwicid:
> Hi @drwicid Thanks for your thorough assessment. I stand corrected :-)
> 
> Please note that the tests are failing. Would you please double check the new tests for `copy` and make sure they reflect the correct intention of the module? Thanks

All tests executed clean today.  Test cases are adjusted to cover force=false with a new file and existing file.

russoz:
@drwicid thanks for the bug fix
@Jonny007-MKD thanks for reviewing (in the issue)

PR #97, opened 2026-01-19
test(integration/copy): check behaviour when force=false

russoz:
##### SUMMARY

The module `copy` has a default `force=true`, which suceeds regardless of the file existing and/or the content being different.

Prompted by #94 and #95, this PR ensures that the module copy is executed with `force=false`:
1. and the content has changed, it pops an error
2. and the content has NOT changed, the task `is not changed`

##### ISSUE TYPE

- Test Pull Request

##### COMPONENT NAME

copy

### DIFF
--- a/tests/integration/targets/copy/tasks/main.yml
+++ b/tests/integration/targets/copy/tasks/main.yml
@@ -60,7 +60,7 @@
   ansible.builtin.assert:
     that:
       - copy_out3c is not changed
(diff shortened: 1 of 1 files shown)


russoz:
The case 2 in the description is not working - that's the bug reported in #94 . When that issue is resolved, we must remove the `ignore_error: true` lines from the test.

PR #116, opened 2026-01-29
Always set openwrt_package_manager

russoz:
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

roles/init/tasks/main.yml


russoz:
Fixes problem noted in #109 by @egandro 

Cc: @Jonny007-MKD

PR #125, opened 2026-02-02
test: add molecule tests for existing modules

zeerayne:
##### SUMMARY

* Adds molecule tests for already implemented modules
* Changes docker container start command to `/sbin/init` to enable full openwrt initialization to be completed (required for ubus-dependant modules to be tested)

##### ISSUE TYPE
- Feature Pull Request

##### ADDITIONAL INFORMATION

In this PR some tests for existing modules were added:
* UCI: write a value
* Service: enable/disable/start/stop
* Sysctl: setting/removing value

For `service` module test to be comleted successfully `ubus` should be operational. Otherwise there is no chance any service can be started. So, PR changes `sleep infinity` container command to `/sbin/init` to enable fully operational OpenWRT to run inside container.

Molecule tests can be run as usual, but more tests are now in scenario
```sh
molecule test

...


PLAY RECAP *********************************************************************
openwrt_x86_64-24.10.4     : ok=60   changed=18   unreachable=0    failed=0    skipped=12   rescued=0    ignored=0
```

### DIFF
--- a/extensions/molecule/default/molecule.yml
+++ b/extensions/molecule/default/molecule.yml
@@ -7,7 +7,7 @@ driver:
 platforms:
   - name: openwrt_x86_64-${OPENWRT_VERSION:-24.10.4}
     image: openwrt/rootfs:x86_64-${OPENWRT_VERSION:-24.10.4}
-    command: /sbin/init # required for proper ubus initialization
+    command: /sbin/init
     pre_build_image: true
 scenario:
   test_sequence:


extensions/molecule/default/converge.yml
```diff
@@ -193,3 +218,132 @@
...
+            enabled: true
+
+        - name: Check cron enabled  # noqa: command-instead-of-module
+          community.openwrt.command:
+            cmd: service cron enabled
+          changed_when: false
+          failed_when: false
+          register: cron_enabled
+
+        - name: Assert cron service status
```
Jonny007-MKD: I have no experience in testing ansible collections. It is verbose as almost everything in ansible :)
Is it uncommon to put the assertion into the `failed_when`?

extensions/molecule/default/converge.yml
```diff
@@ -193,3 +218,132 @@
...
+          community.openwrt.sysctl:
+            sysctl_file: "{{ sysctl_file }}"
+            state: present
+            name: "{{ sysctl_setting_name }}"
+            value: "{{ sysctl_setting_value }}"
+
+        - name: Check sysctl setting exists
+          community.openwrt.lineinfile:
+            path: "{{ sysctl_file }}"
+            line: "{{ sysctl_setting_name + '=' + sysctl_setting_value | string }}"
```
Jonny007-MKD:
```suggestion
            line: "{{ sysctl_setting_name }}={{ sysctl_setting_value }}"
```

extensions/molecule/default/converge.yml
zeerayne: Not sure how to combine assertion with `failed_when`, I can change is to
```yaml
  ignore_errors: true
  register: cron_enabled

- name: Assert cron service status
  ansible.builtin.assert:
    that:
      - cron_enabled | failed
```
which is mostly similar to what it is now

extensions/molecule/default/molecule.yml
```diff
@@ -7,7 +7,7 @@ driver:
 platforms:
   - name: openwrt_x86_64-${OPENWRT_VERSION:-24.10.4}
     image: openwrt/rootfs:x86_64-${OPENWRT_VERSION:-24.10.4}
-    command: sh -c "mkdir -p /var/lock; sleep infinity"
+    command: /sbin/init # required for proper ubus initialization
```
russoz: Is that affecting any of the tests? When you say "proper ubus initialization", does it mean just "starting ubus" or is there something else? What value we get from doing it, that we are missing today?

And for the record, we have other molecule scenarios that could/should be updated as well, if we agree on adding this.

russoz (commented):
Hi @zeerayne thanks for your contribution!

I have a couple of comments/questions.

extensions/molecule/default/molecule.yml
russoz: Err, sorry, I read the description again and the explanation is right there :-)

In that case, please make this change also in the scenarios:

- integration_test
- role_init_install_recommended_false
- role_init_install_recommended_true

extensions/molecule/default/converge.yml
```diff
@@ -54,6 +54,31 @@
...
+    - name: Read written value UCI
+      community.openwrt.uci:
+        command: get
+        key: dhcp.@dnsmasq[0].domain
+      register: uci_dhcp_dnsmasq_domain
+
+    - name: Assert domain value (UCI)
+      ansible.builtin.assert:
+        that:
+          - uci_dhcp_dnsmasq_domain.result == "test"
```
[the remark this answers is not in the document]
zeerayne: Looks like my UCI tests are redundant and mostly duplicate an existing ones, will drop it then

extensions/molecule/default/converge.yml
```diff
@@ -193,3 +193,67 @@
...
+            path: /etc/crontabs/root
+            state: touch
+            content: |
+              * * * * * /bin/echo "Holy canole,    it\\'s {{ os_release.name }}"
+
+        - name: Disable cron service
+          community.openwrt.service:
+            name: cron
+            enabled: false
+
```
russoz: And this would probably be better placed in the integration test for service as well.

zeerayne: Moved this one to integration tests for `service` module

russoz:
@zeerayne Thanks for your contribution!
@Jonny007-MKD Thanks for reviewing!

Issue #63, opened 2025-12-26
community.openwrt.fetch & community.openwrt.template don't exist


galliangg:
##### SUMMARY
With gekmihesg.openwrt I was able to use fetch and template.
Is it possible to create those commands here?

##### ISSUE TYPE
- Feature Idea

##### ADDITIONAL INFORMATION

```yaml

```

russoz:
Hi @galliangg ,

I don't see why not, just bear in mind right now we are prioritising getting the collection ready for a 1.0.0 release, and things like tests for modules are missing yet.

That being said, if you want to risk writing those modules yourself, we would be happy to help.

russoz:
And I would risk saying that they might end up being implemented as action plugins rather than shell modules.

zeerayne:
Posting here simple template workaround using `community.openwrt.copy`.
This achieves similar results as if `ansible.builtin.template` was used.

```yaml
    - name: Template workaround
      community.openwrt.copy:
        content: "{{ lookup('template', 'actual_template.j2') }}"
        dest: /etc/app/config.json
```

tfasz:
> Posting here simple template workaround using `community.openwrt.copy`. This achieves similar results as if `ansible.builtin.template` was used.
> 
>     - name: Template workaround
>       community.openwrt.copy:
>         content: "{{ lookup('template', 'actual_template.j2') }}"
>         dest: /etc/app/config.json

Nice trick @zeerayne - that's easier than what I had done for a work-around. I'll add that as an example to the docs.

russoz:
Not doing this right now, but I reckon we should break this into two issues (or more, if we have more modules we want to implement), so that they can be closed when implemented. `template` already has a candidate PR (in early stages, but nonetheless)

PR #137, opened 2026-02-10
feat: `template` module

zeerayne:
##### SUMMARY

This PR implements `template` module.
Relates to https://github.com/ansible-collections/community.openwrt/issues/63

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

This is a `community.openwrt.template`

##### ADDITIONAL INFORMATION

This is mostly a copy of built-in `template`  module. Code taken from here:
https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/action/template.py

I've tried to figure out if I can change [`ansible.legacy.copy`](https://github.com/ansible/ansible/blob/cf28bcb65d2b280607ea174c58aa42c9f873f421/lib/ansible/plugins/action/template.py#L168) to `community.openwrt.copy` without copy-pasting the full module, but I failed to find a good way to do this.

So, I decided to copy existing template module and adapt it to use `community.openwrt.copy` for transferring files to remote host.

### DIFF
--- a/plugins/action/template.py
+++ b/plugins/action/template.py
@@ -1,70 +1,32 @@
 # Copyright: (c) 2026, Ilya Bogdanov (@zeerayne)
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 # SPDX-License-Identifier: GPL-3.0-or-later
-import ast
-import inspect
-import textwrap
 
+from unittest import mock
+
+from ansible.plugins.loader import action_loader
 from ansible.plugins.action.template import ActionModule as TemplateActionModule
 
 
-def patch_and_create_class(original_cls):
-    source = textwrap.dedent(inspect.getsource(original_cls))
-    tree = ast.parse(source)
+class ActionModule(TemplateActionModule):
+    def run(self, tmp=None, task_vars=None):
 
-    class AnsibleActionTransformer(ast.NodeTransformer):
-        def visit_Assign(self, node):
-            # Find line: copy_action = self._shared_loader_obj.action_loader.get('ansible.legacy.copy', ...)
-            # Check that this is an assignment to the variable copy_action
-            if (isinstance(node.targets[0], ast.Name) and node.targets[0].id == 'copy_action'):
+        _getter = self._shared_loader_obj.action_loader.get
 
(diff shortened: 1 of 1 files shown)


russoz:
Hi @zeerayne sorry for the delay, couple of busy days around here.

Err, I also do not like the idea of just copying the thing over, but also not sure we can do much better than that. The templating is meant to be Jinja2, which necessarily implies being Python, which, in c.owrt, necessarily implies you can only do it in the action plugin, not in the module (/bin/sh) code.

Could you please pinpoint in the code what you changed, compared to the original? TIA

zeerayne:
> Wow, I've never seen AST stuff with Python before! What do you think about its stability? Is it going to work as long as there is such a call to `get` with this argument?

Well, It will return the untouched class if `copy_action = ` line is not present in the original source.

So yes, it can break in the future if core ansible template action will be refactored because I'm relying on ansible's source code. But looks like this is more stable approach than copying the whole class (which will also break if ansible's internal structure changes, like in transit 2.18 -> 2.19)

russoz (commented):
Hi @zeerayne, thanks for the update! Wow, the AST feat is indeed impressive. I am not sure we _should_ do that though, but I do not have another solution to offer at this point.

I think I will leave this PR open until a couple of days before the 1.1.0 release (less than 2 weeks time), to give time for more people to see it and possibly comment. If no one brings any improvements or alternatives, this should make it to community.openwrt 1.1.0.

russoz:
@zeerayne please note that this PR now has conflicts. You must rebase from `main` and fix the conflict.

Instructions for rebasing: https://docs.ansible.com/ansible/latest/dev_guide/developing_rebasing.html

russoz:
@zeerayne please note that this PR now has conflicts. You must rebase from `main` and fix the conflict.

No pressure, just be aware that we should release 1.1.0 this week - it was meant to be tomorrow, but I will extend the deadline to Friday on the hope that OpenWrt 25.12.0 is already released.

Instructions for rebasing: https://docs.ansible.com/ansible/latest/dev_guide/developing_rebasing.html

zeerayne:
@russoz looks like I have managed to drop merge commit from branch and preserve all other stuff

russoz:
@zeerayne thanks for your contribution!!
@Jonny007-MKD thanks for reviewing!!

PR #138, opened 2026-02-11
feat(modules/setup): collect date & time facts

### LINKED ISSUE #52: `community.openwrt.setup` does not set the date/time facts
> to be fair, `community.openwrt.setup` also does not set the date/time facts - and it does not look very complicated to do, so it is definitely one for the future improvements

_Originally posted by @russoz in https://github.com/ansible-collections/community.openwrt/pull/39#discussion_r2637362777_

russoz:
##### SUMMARY

Collect and return date & time facts in the same structure as `ansible.builtin.setup.ansible_date_time`.

Fixes #52 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

setup

### DIFF
--- a/changelogs/fragments/138-setup-date.yml
+++ b/changelogs/fragments/138-setup-date.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - setup - collect ``ansible_date_time`` facts as well (https://github.com/ansible-collections/community.openwrt/issues/52, https://github.com/ansible-collections/community.openwrt/pull/138)`
+  - setup - collect ``ansible_date_time`` facts as well (https://github.com/ansible-collections/community.openwrt/issues/52, https://github.com/ansible-collections/community.openwrt/pull/138).


plugins/modules/setup.sh
```diff
@@ -3,7 +3,57 @@
...
+    epoch_int=$(date +%s -d "@$now")
+    date=$(date +%Y-%m-%d)
+    time=$(date +%H:%M:%S)
+    iso8601_micro=$(date +%Y-%m-%dT%H:%M:%S.000000Z "@$utcnow")
+    iso8601=$(date +%Y-%m-%dT%H:%M:%SZ "@$utcnow")
+    iso8601_basic=$(date +%Y%m%dT%H%M%S000000 "@$now")
+    iso8601_basic_short=$(date +%Y%m%dT%H%M%S "@$now")
+    tz=$(date +%Z)
+    tz_dst=$(date +%Z)
+    tz_offset=$(date +%z)
```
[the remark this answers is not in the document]
russoz: Hi @Jonny007-MKD, thanks for reviewing.

This is a very clever way of optimising the `date` commands, and I am happy to accept the suggestion.

That being said, if we need 500 iterations to produce a difference slightly longer than 15s => ~0.03s per iteration (crude simplification) for something that is executed only once during each playbook's lifecycle, I would argue that the impact is negligible.

And as of the comment you left about using local time or UTC time, those date statements were adapted from the original code at:
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/facts/system/date_time.py#L43-L68

You will notice:
```python
        date_time_facts['iso8601_micro'] = utcnow.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
        date_time_facts['iso8601'] = utcnow.strftime("%Y-%m-%dT%H:%M:%SZ")
        date_time_facts['iso8601_basic'] = now.strftime("%Y%m%dT%H%M%S%f")
        date_time_facts['iso8601_basic_short'] = now.strftime("%Y%m%dT%H%M%S")
```

That the `iso8601_basic*` facts use localtime, whereas the other `iso8601*` facts use UTC time. So your suggestion is in need of some adjusting.

russoz:
Anyways, the CI tests will fail because the apk repository (for OpenWrt 25.12.0-rc4) is not working at this moment. 

See https://github.com/openwrt/openwrt/issues/21981 for details.

plugins/modules/setup.sh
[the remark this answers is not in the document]
Jonny007-MKD: I agree that the performance gain is not huge :) Certainly it takes longer on older/smaller (i.e. slower) devices, so it could help. Anyway, I had the idea that starting date 18 times could be improved, and I wanted to see how big the impact is ;) And I'm a bit annoyed of how long it takes ansible to run moderate complex playbooks.

In my first draft I had
```
    iso8601="$uyear-$umonth-$uday"T"$uhour:$uminute:$usecond"Z
```
but it yielded different results than yours on my router; so I removed all the `u`s. I hope you can sort that out :)

russoz:
hi @Jonny007-MKD, I gave it some a spin, also added support for microseconds (if coreutils-date is installed).

tests/integration/targets/setup/tasks/main.yml
```diff
@@ -17,3 +21,11 @@
   ansible.builtin.assert:
     that:
       - setup_result.ansible_facts.ansible_os_family is defined
+
+- name: Assert date/time facts
+  vars:
+    dt: "{{ setup_result.ansible_facts.ansible_date_time | default({}) }}"
+  ansible.builtin.assert:
+    that:
+      - dt is defined
```
Jonny007-MKD: isn't this always true as you default it?

tests/integration/targets/setup/tasks/main.yml
```diff
@@ -17,3 +21,16 @@
...
+
+- name: Assert date/time facts
+  vars:
+    dt: "{{ setup_result.ansible_facts.ansible_date_time }}"
+  ansible.builtin.assert:
+    that:
+      - dt is defined
+      - dt.date == "{{ dt.year }}-{{ dt.month }}-{{ dt.day }}"
+      - dt.time == "{{ dt.hour }}-{{ dt.minute }}-{{ second }}"
+      - dt.iso8601_basic_short == {{ dt.year }}-{{ dt.month }}-{{ dt.day }}T{{ dt.hour }}-{{ dt.minute }}-{{ second }}"
```
Jonny007-MKD: my bad

```suggestion
      - dt.time == "{{ dt.hour }}-{{ dt.minute }}-{{ dt.second }}"
      - dt.iso8601_basic_short == {{ dt.year }}-{{ dt.month }}-{{ dt.day }}T{{ dt.hour }}-{{ dt.minute }}-{{ dt.second }}"
```

russoz: All good 😄 we have all been there before

PR #159, opened 2026-03-02
test(integration): molecule-only integration tests

### LINKED ISSUE #147: Improve Integration Test mechanism
##### SUMMARY

The original idea of having the integration tests executed with `ansible-test integration` was implemented in #49 and it has, to the extent imagined then, worked.

However, as the collection is growing and more contributions are arriving and the tests are being executed more frequently we have started to bump into issues. As the tests are now, in every single run:

* For each target (distinct OpenWrt version):
  * For every single module (integration test target), the test script will:
    * Install supporting collections (ansible-galaxy -> galaxy.ansible.com)
    * Install molecule and dependencies (pip -> pypi)
    * Install docker (apt -> apt repos)
    * Create the container for the test, destroy the container for the test (molecule -> docker hub)

This has many negative consequences to the testing:
 
* Repeatedly relying on external dependencies during the test
* The added complexity is not bringing any return at the moment
* On occasion, I had three or four PRs with tests running at the same time, eventually I started getting errors of rate limits from docker hub
* The sheer amount of logs makes troubleshooting more painful than it had to be
* The tests take longer (because of all the installing/creating/destroying)
* The tests in general have been very flaky, and many times the re-run goes successful without any changes in the code

The proposal in this issue is to:

* Make the integration tests run straight from molecule
  * The tests would be executed for all targets, always, however ...
  * ... some targets will require running on pristine containers, whilst others can be executed in the same container run: the isolated ones should have distinct molecule scenarios, the others should be in one single molecule scenario
* The integration tests should not be dismantled - they should not run in the CI, but they would be useful for running local tests.

The expected benefits are:

* Faster test times
* Easier to trouble shoot
* Reduced dependency on external services during test time

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

NA

russoz:
##### SUMMARY

Replace integration tests with `ansible-test integration` with direct `molecule`.

Fixes #147 

##### ISSUE TYPE

- Test Pull Request

##### COMPONENT NAME

NA

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -171,7 +171,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           path: ansible_collections/community/openwrt
 


russoz:
one more round

russoz:
success again

russoz:
success again

extensions/molecule/molintegration/converge.yml
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) 2026, Alexei Znamensky (@russoz)
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+- name: Converge
+  hosts: all
+  gather_facts: false
+  vars:
+    integration_tests_targets:
```
zeerayne: Looks like those vars can be fetched dynamically with smth like that

```yaml
    - name: Find all test targets
      ansible.builtin.find:
        paths: "tests/integration/targets"
        file_type: directory
        recurse: no
      register: test_targets_dirs

    - name: Set `integration_tests_targets` fact
      ansible.builtin.set_fact:
        integration_tests_targets: "{{ test_targets_dirs.files | map(attribute='path') | map('basename') | list }}"
```

russoz: Ha, and that was the reason: `ansible.builtin.find` will try to run Python in the remote target, and that will fail. We cannot use that module here.

That explains why I have a local branch to implement a new module `community.openwrt.find`, but I wanted this PR to go in as soon as possible, so I settled with a fixed list until we have `c.o.find` available.

PR #211, opened 2026-04-03
fix(modules/setup): filter out wifi secrets

### LINKED ISSUE #38: setup: wifi password is leaking in the facts output
##### SUMMARY

The passwords for the configured SSIDs are leaked in the output of the `setup` module.

Maybe it would be filtered by default, and an extra parameter would allow it to be returned.

I presume this would be more easily handled in the action plugin, but not sure of the implications here.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

setup

##### ANSIBLE VERSION
NA

##### COLLECTION VERSION
NA

##### CONFIGURATION

```paste below

```

##### STEPS TO REPRODUCE

```yaml
---
- name: Full config
  hosts: router_group
  become: false
  gather_facts: false
  tasks:
    - name: Init community.openwrt
      vars:
        openwrt_install_recommended_packages: true
      ansible.builtin.import_role:
        name: community.openwrt.init

    - name: Collect facts
      community.openwrt.setup:
      register: owrt_facts

    - name: Print facts
      ansible.builtin.debug:
        var: owrt_facts
```

##### EXPECTED RESULTS

`openwrt_wireless.radioX.interfaces[].config.key` would not contain a wifi password.

##### ACTUAL RESULTS
`openwrt_wireless.radioX.interfaces[].config.key` contains a wifi password.

russoz:
## SUMMARY

Filter out secrets from the wifi configuration in the output of `community.openwrt.setup`.

Fixes #38 

## ISSUE TYPE

- Bugfix Pull Request

## COMPONENT NAME

plugins/action/setup.py
plugins/modules/setup.py

### DIFF
--- a/plugins/action/setup.py
+++ b/plugins/action/setup.py
@@ -22,6 +22,8 @@ _WIRELESS_SENSITIVE_KEYS = frozenset(
         "priv_key2_pwd",  # EAP private key passphrases
         "private_key_passwd",  # hostapd server private key passphrase
         "multi_ap_backhaul_key",  # Multi-AP backhaul key
+        "r0kh",  # 802.11r roaming key holder 0 (contains shared keys)
+        "r1kh",  # 802.11r roaming key holder 1 (contains shared keys)
     }
 )
 
--- a/plugins/action/setup.py
+++ b/plugins/action/setup.py
@@ -6,13 +6,20 @@ from __future__ import annotations
 
 from ansible_collections.community.openwrt.plugins.plugin_utils.openwrt_action import OpenwrtActionBase
 
-_WIRELESS_SENSITIVE_KEYS = frozenset({
-    "key", "key1", "key2", "key3", "key4",  # WPA PSK / WEP
-    "sae_password",                          # WPA3-SAE
-    "password",                              # EAP user password
-    "auth_secret", "acct_secret",            # RADIUS shared secrets
-    "priv_key_pwd",                          # EAP private key password
-})
+_WIRELESS_SENSITIVE_KEYS = frozenset(
+    {
+        "key",
+        "key1",
+        "key2",
+        "key3",
+        "key4",  # WPA PSK / WEP
+        "sae_password",  # WPA3-SAE
+        "password",  # EAP user password
+        "auth_secret",
+        "acct_secret",  # RADIUS shared secrets
+        "priv_key_pwd",  # EAP private key password
+    }
+)
 
 
 def _redact_wireless(obj):
--- a/plugins/action/setup.py
+++ b/plugins/action/setup.py
@@ -6,7 +6,13 @@ from __future__ import annotations
 
 from ansible_collections.community.openwrt.plugins.plugin_utils.openwrt_action import OpenwrtActionBase
 
-_WIRELESS_SENSITIVE_KEYS = frozenset({"key", "password", "psk"})
+_WIRELESS_SENSITIVE_KEYS = frozenset({
+    "key", "key1", "key2", "key3", "key4",  # WPA PSK / WEP
+    "sae_password",                          # WPA3-SAE
+    "password",                              # EAP user password
+    "auth_secret", "acct_secret",            # RADIUS shared secrets
+    "priv_key_pwd",                          # EAP private key password
+})
 
 
 def _redact_wireless(obj):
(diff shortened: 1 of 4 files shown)


plugins/action/setup.py
```diff
@@ -6,6 +6,36 @@
...
+
 
 class ActionModule(OpenwrtActionBase):
-    pass
+    def run(self, tmp=None, task_vars=None):
+        result = super().run(tmp, task_vars)
+
+        ansible_facts = result.get("ansible_facts", {})
+        if "openwrt_wireless" in ansible_facts:
+            ansible_facts["openwrt_wireless"] = _redact_wireless(ansible_facts["openwrt_wireless"])
```
zeerayne: Maybe to make this behaviour controlled? Some fact like `openwrt_install_recommended_packages`? By default sensitive facts should be redacted, but for some special cases it can be turned off.

russoz: That was my first approach. But then I started wondering why and how easy/hard this _should_ be. The information is there, accessible, to anyone who have root access to the routers. Nothing that a `command` and a  `community.general.json_query()` cannot solve. 

Mixed feelings.

zeerayne: Agreed, no questions left here

russoz:
Planning to merge this in a couple of days unless anyone objects.

plugins/action/setup.py
```diff
@@ -6,6 +6,36 @@
...
+        "key1",
+        "key2",
+        "key3",
+        "key4",  # WPA PSK / WEP
+        "sae_password",  # WPA3-SAE
+        "password",  # EAP user password
+        "auth_secret",
+        "acct_secret",  # RADIUS shared secrets
+        "priv_key_pwd",  # EAP private key password
+    }
```
Jonny007-MKD, line 27: I'd also add [private_key_passwd](https://git.openwrt.org/openwrt/openwrt/tree/package/network/config/wifi-scripts/files/lib/netifd/hostapd.sh?id=df338d67d4afe4aa1819b2f04a442b84c75ea8c9#n396)

Jonny007-MKD (approved):

plugins/action/setup.py
```diff
@@ -6,6 +6,36 @@
...
+        "key",
+        "key1",
+        "key2",
+        "key3",
+        "key4",  # WPA PSK / WEP
+        "sae_password",  # WPA3-SAE
+        "password",  # EAP user password
+        "auth_secret",
+        "acct_secret",  # RADIUS shared secrets
+        "priv_key_pwd",  # EAP private key password
```
sguarin: I would add mobility_domain, I think it's sensitive.

plugins/action/setup.py
sguarin: I like the idea of the boolean to enable sensitive information extraction.

sguarin: (if someone was using the key/passwords reporting, he can easily fix the change of behaviour by one boolean).

russoz: Not a big fan of the idea TBH, but I reckon that if we don't add it, I'll spend the rest of my life responding why. However, I will insist on making the default NOT to expose the secrets.

plugins/action/setup.py
russoz: Hi @sguarin I failed to find information related to `mobility_domain` being an attack surface. Could you please expand on why you deem it sensitive? TIA

plugins/action/setup.py
russoz: done

plugins/action/setup.py
russoz: done, added a `expose_secrets` param, but its default is `false` (breaking change).

Jonny007-MKD (approved):

plugins/action/setup.py
sguarin: It´s about 802.11r (roaming)... if not set it's derived from ssid/password, anyways the surface attack is no big deal because the attacker needs to have network access in order to publish it's a member of the roaming network.
Another two I just found looking for references are: `r0kh` and `r1kh`.

Reference:
https://openwrt.org/docs/guide-user/network/wifi/basic

russoz: That page says that its default value is

> (first 4 bytes of md5sum of the SSID)

My cryptography-fu is not that sharp, but I would be extremely surprised if those 4 bytes could be used for _anything_ at all.

OTOH, `r0kh` and `r1kh` seem to contain a _key_ value in its text. So they should definitely go in.

Jonny007-MKD (approved):

russoz:
Ok, looks like we have contented crowd. I am planning on merging this in the next couple of days.

russoz:
@Jonny007-MKD @sguarin @zeerayne thanks for reviewing!!

PR #216, opened 2026-04-07
feat(action): add community.openwrt.wait_for_connection

russoz:
## SUMMARY

Adds `community.openwrt.wait_for_connection`, which wraps the upstream `ansible.builtin.wait_for_connection` action plugin but redirects its internal `ansible.legacy.ping` call to `community.openwrt.ping` (shell-based). This avoids the `python3 not found` failure that occurs on OpenWrt devices, which have no Python installed.

The redirect is implemented using `mock.patch.object` on `_execute_module`, following the same pattern used by `community.openwrt.template`.

Also extracts a `_run_shell_module()` helper on `OpenwrtActionBase` to avoid duplication between action plugins.

## ISSUE TYPE

## COMPONENT NAME

wait_for_connection

## ADDITIONAL INFORMATION

The handler in `roles/common/handlers/main.yml` still calls `ansible.builtin.wait_for_connection`. A follow-up PR will update it to use `community.openwrt.wait_for_connection`.

### DIFF
--- a/plugins/action/wait_for_connection.py
+++ b/plugins/action/wait_for_connection.py
@@ -1,0 +1,24 @@
+# Copyright (c) 2026 Alexei Znamensky
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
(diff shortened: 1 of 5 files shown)


russoz:
Rebased after #217 was merged

PR #221, opened 2026-04-22
Maintenance process

russoz:
## SUMMARY

Changes to the Maintenance process, specially the release cadence and deprecation.

As we created this collection, most of the processes were copied over from `commmunity.general`, but as we move forward it looks like they could use some adjustment.

### Release Cadence

The proposal is to release major versions after OpenWrt releases (e.g. 25.12, 26.xx, 27.yy, etc...). I am suggesting a lag after the release to account for adjustments or last-minute bugfixes. The major drive for changes in this collection are the OpenWrt releases. Trying to forcibly fit a scheduled release on that feels awkward. For example, I reckon OpenWrt 25.12 should have warranted a major release.

On the other hand, we propose a tentative release every couple of weeks: 

- If we had no PRs, then skip the release
- If we had only bugfixes, then it is a patch/bugfix release: x.y.z -> x.y.(z+1)
- If we had new features (regardless of bugfixes), then it is a minor release: x.y.z -> x.(y+1).z

These fortnightly releases would be a perfect match for [Semantic Releases](https://python-semantic-release.readthedocs.io/en/latest/index.html), but I reckon we should do that one step at a time. After agreeing (hopefully) on the cadence, then we could adopt conventional commits as a standard, eventually enforcing it, and from there onward it would be a matter integrating Semantic Releases in the release process, keeping the Changelog mechanism used in Ansible Community collections.

After the release yesterday (20-21/Apr), we now have roughly four weeks to debate on this. If this is agreed upon way before that, we may cancel the release 1.4.0 in May and start the fortnightly releases. If we don´t agree or if we decide not to pursue this, then we release 1.4.0 as expected and take it from there.

### Deprecation

We have not had any deprecations yet, but we already have one in scope: when OpenWrt 24.xx falls out of support, we will deprecate `community.openwrt.opkg`. 

Deprecated features will only be removed in major versions, but as we are proposing to lean on OpenWrt release schedule, we would not have a fixed period for that. Therefore, the plan would be to remove the feature in the first major version that takes place after 12 months of the deprecation. That way users will have plenty of time to adjust their code, if needed.

## ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/MAINTENANCE.md
+++ b/MAINTENANCE.md
@@ -50,7 +50,7 @@ The collection will release minor versions periodically, between major versions.
 
 - Patch versions `x.y.z` until the last minor release of a major release branch will only be released when necessary. The intended frequency is _never_, they are reserved for packaging failures, or fixing major breakage / security problems.
 
-### Deprecation policy
+## Deprecation policy
 
 - Deprecations are done by version number (not by date).
 - New deprecations can be added during every minor release, under the condition that they **do not break backwards compatibility**.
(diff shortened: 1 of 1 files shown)


russoz:
I am assuming there are no objections.

Issue #158, opened 2026-02-27
Salvage issues not fixed from gekmihesg.openwrt


russoz:
##### SUMMARY

The https://github.com/gekmihesg/ansible-openwrt/ repository was archived on Feb 20, 2026, closing all the pending issues.

However, some of those issue might still apply to community.openwrt. The purpose of this ticket is to scoop those closed issues from:
https://github.com/gekmihesg/ansible-openwrt/issues?q=is%3Aissue%20state%3Aclosed

and open new issues when applicable/relevant.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

NA

PR #224, opened 2026-04-25
feature(apk): Add option allow_untrusted to be able to install local packages (#195)

### LINKED ISSUE #195: opkg/apk: enable installation from local file
##### SUMMARY

As of now, both `opkg` and `apk` modules do not support installing a package directly from a file.

This was previously reported (in a slightly different perspective) in
https://github.com/gekmihesg/ansible-openwrt/issues/70 

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

apk
opkg

##### ADDITIONAL INFORMATION

```yaml

```

Jonny007-MKD:
## SUMMARY
Add feature to `apk` to install local packages by adding the `--allow-untrusted` option

Fixes #195

## ISSUE TYPE
- Feature Pull Request

## COMPONENT NAME

plugins/modules/apk.py
plugins/modules/apk.sh

## ADDITIONAL INFORMATION
We certainly should add tests for it as well. Almighty AI says building APKs is quite easy:
- `install alpine-sdk`
- `mkdir -p /tmp/packages/testpkg`
- with workdir `/tmp/packages/testpkg`: `abuild init`
- write to `/tmp/packages/testpkg/APKBUILD`:
```
# Maintainer: Test <test@example.com>
pkgname=testpkg
pkgver=1.0
pkgrel=0
pkgdesc="Minimal test package"
url="https://example.com"
arch="all"
license="MIT"
options="!check"

package() {
    install -Dm644 /dev/null "$pkgdir"/usr/share/testpkg/placeholder
}
```
- with workdir `/tmp/packages/testpkg`: `abuild -r` -> creates `/tmp/packages/testpkg/pkg/testpkg-1.0-r0.apk`

Alternatively we could do it by hand:
- Write to `/tmp/testpkg/.PKGINFO`:
```
pkgname = testpkg
pkgver = 1.0-r0
arch = x86_64
license = MIT
```
- `tar -cvzf testpkg-1.0-r0.apk -C /tmp/testpkg /tmp/`

I'd go with the 2nd option? wdyt?

I haven't set up a local test environment, yet :)

### DIFF
--- a/changelogs/fragments/220-apk-allow-untrusted.yml
+++ b/changelogs/fragments/220-apk-allow-untrusted.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "apk - add option ``allow_untrusted`` to be able to install local packages (https://github.com/ansible-collections/community.openwrt/issues/195, https://github.com/ansible-collections/community.openwrt/pull/224)"
--- a/plugins/modules/apk.py
+++ b/plugins/modules/apk.py
@@ -55,6 +55,11 @@ options:
       - Continue even if package dependencies are broken.
     type: bool
     default: false
+  allow_untrusted:
+    description:
+      - To install a locally available apk package, for example if this device has no internet access but you can upload apk packages directly to it.
+    type: bool
+    default: false
 """
 
 EXAMPLES = r"""
@@ -79,6 +84,12 @@ EXAMPLES = r"""
     name: tcpdump
     state: present
     no_cache: true
+
+- name: Install local package
+  community.openwrt.apk:
+    name: /path/to/file.apk
+    state: present
+    allow_untrusted: true
 """
 
 RETURN = r""""""
--- a/plugins/modules/apk.py
+++ b/plugins/modules/apk.py
@@ -55,6 +55,12 @@ options:
       - Continue even if package dependencies are broken.
     type: bool
     default: false
+  allow_untrusted:
+    description:
+      - To install a locally available apk package, for example if this device has no internet access but you can upload apk packages directly to it.
+    type: bool
+    default: false
+    version_added: 1.5.0
 """
 
 EXAMPLES = r"""
@@ -79,6 +85,12 @@ EXAMPLES = r"""
     name: tcpdump
     state: present
     no_cache: true
+
+- name: Install local package
+  community.openwrt.apk:
+    name: /path/to/file.apk
+    state: present
+    allow_untrusted: true
 """
 
 RETURN = r""""""
--- a/plugins/modules/apk.sh
+++ b/plugins/modules/apk.sh
@@ -20,7 +20,7 @@ install_packages() {
     [ -n "$pkgs_to_install" ] || return 0
 
     [ -n "$_ansible_check_mode" ] || {
-        apk $update_cache $no_cache $force_broken_world add $pkgs_to_install >"$out" 2>"$err"
+        apk $update_cache $no_cache $force_broken_world $allow_untrusted add $pkgs_to_install >"$out" 2>"$err"
         rc=$?
         stdout="$(cat "$out")"
         stderr="$(cat "$err")"
@@ -62,6 +62,7 @@ init() {
         update_cache/bool
         no_cache/bool
         force_broken_world/bool
+        allow_untrusted/bool
     "
     RESPONSE_VARS="
         stdout/str/a
@@ -91,6 +92,7 @@ main() {
     [ -z "$update_cache" ] || update_cache="--update-cache"
     [ -z "$no_cache" ] || no_cache="--no-cache"
     [ -z "$force_broken_world" ] || force_broken_world="--force-broken-world"
+    [ -z "$allow_untrusted" ] || allow_untrusted="--allow-untrusted"
 
     case "$state" in
         present|installed) install_packages;;
--- a/plugins/modules/apk.sh
+++ b/plugins/modules/apk.sh
@@ -20,12 +20,18 @@ install_packages() {
     [ -n "$pkgs_to_install" ] || return 0
 
     [ -n "$_ansible_check_mode" ] || {
-        apk $update_cache $no_cache $force_broken_world add $pkgs_to_install >"$out" 2>"$err"
+        apk $update_cache $no_cache $force_broken_world $allow_untrusted add $pkgs_to_install >"$out" 2>"$err"
         rc=$?
         stdout="$(cat "$out")"
         stderr="$(cat "$err")"
         # Verify installation
         for pkg in $pkgs_to_install; do
+            if [ -f "$pkg" ]; then
+                # we installed a file and have to query the file for the package name to verify its installation
+                json_load "$(apk adbdump --format json "$pkg")" || fail "could not parse output of apk adbdump"
+                json_select "info"
+                json_get_var pkg "name"
+            fi
             query_package "$pkg" || fail "failed to install $pkg: $stdout $stderr"
         done
     }
@@ -62,6 +68,7 @@ init() {
         update_cache/bool
         no_cache/bool
         force_broken_world/bool
+        allow_untrusted/bool
     "
     RESPONSE_VARS="
         stdout/str/a
@@ -91,6 +98,7 @@ main() {
     [ -z "$update_cache" ] || update_cache="--update-cache"
     [ -z "$no_cache" ] || no_cache="--no-cache"
     [ -z "$force_broken_world" ] || force_broken_world="--force-broken-world"
+    [ -z "$allow_untrusted" ] || allow_untrusted="--allow-untrusted"
 
     case "$state" in
         present|installed) install_packages;;
--- a/tests/integration/targets/apk/tasks/main.yml
+++ b/tests/integration/targets/apk/tasks/main.yml
@@ -37,3 +37,27 @@
       ansible.builtin.assert:
         that:
           - apk_remove is succeeded
+
+    - name: Test --allow-untrusted
+      block:
+        - name: Download a test package
+          ansible.builtin.command:
+            cmd: 'apk fetch htop -s > /tmp/htop.apk'
+
+        - name: Install untrusted package
+          community.openwrt.apk:
+            name: /tmp/htop.apk
+            state: present
+            allow_untrusted: true
+
+        - name: "Verify package is installed"
+          community.openwrt.command:
+            cmd: apk info -e htop
+          register: apk_verify_untrusted
+          failed_when: apk_verify_untrusted.rc != 0
+
+      finally:
+        - name: Remove downloaded package
+          ansible.builtin.file:
+            path: /tmp/htop.apk
+            state: absent
--- a/tests/integration/targets/apk/tasks/main.yml
+++ b/tests/integration/targets/apk/tasks/main.yml
@@ -37,3 +37,30 @@
       ansible.builtin.assert:
         that:
           - apk_remove is succeeded
+
+    - name: Test --allow-untrusted
+      block:
+        - name: Download a test package
+          community.openwrt.command:
+            cmd: 'apk fetch terminfo -o {{ remote_tmp_dir }}'
+          register: apk_fetch_result
+
+        - name: Set apk filename
+          ansible.builtin.set_fact:
+            apk_terminfo_filename: "{{ remote_tmp_dir }}/{{ apk_fetch_result.stdout_lines[0].split(' ')[1] }}.apk"
+
+        - name: Output apk filename
+          ansible.builtin.debug:
+            var: apk_terminfo_filename
+
+        - name: Install untrusted package
+          community.openwrt.apk:
+            name: "{{ apk_terminfo_filename }}"
+            state: present
+            allow_untrusted: true
+
+        - name: "Verify package is installed"
+          community.openwrt.command:
+            cmd: apk info -e terminfo
+          register: apk_verify_untrusted
+          failed_when: apk_verify_untrusted.rc != 0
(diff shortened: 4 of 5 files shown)


changelogs/fragments/220-apk-allow-untrusted.yml
```diff
@@ -0,0 +1,2 @@
+features:
+  - "apk - add option `allow_untrusted` to be able to install local packages (https://github.com/ansible-collections/community.openwrt/issues/195)"
```
russoz: `features` is not a valid keyword for the fragment. See changelogs/config.yaml
```suggestion
minor_changes:
  - "apk - add option ``allow_untrusted`` to be able to install local packages (https://github.com/ansible-collections/community.openwrt/issues/195)"
```

It should also include this URL of this PR.

russoz (commented):
Hi @Jonny007-MKD thanks for the PR! 

Other than the changelog fragment in need of adjustment:

- to fix #195 the PR needs to ensure that `opkg` can also install from files - that was the original issue in the role's repo. And we are still supporting `opkg` for a significant length of time.
- test: I believe that _building_ a package for the test might be a bit overkill. Pending some manual verification, I believe that in both apk and opkg we might be able to download a package from the official repos and then trying to install them locally would have the same effect, assuming there is no cryptographic signing in these packages. Else, in the forum.openwrt.org there's always someone announcing their package XYZ for whatever feature, so we might pick one of those and use it for the test.

Jonny007-MKD:
Hi @russoz,

yes, #195 targets both package managers. Either this PR only fixes one half of #195, or we split it into two issues.

I would not add new features for opkg, though. Until now nobody could use this feature and I would not recommend writing playbooks for opkg anymore.

Downloading a package is a good idea. I thought it would still be signed, but a short test shows that it is untrusted after it was downloaded:

```
# apk fetch sane-test
Downloading sane-test-1.4.0-r1
# apk add ./sane-test-1.4.0-r1.apk 
ERROR: ./sane-test-1.4.0-r1.apk: UNTRUSTED signature
# apk add ./sane-test-1.4.0-r1.apk  --allow-untrusted
(1/4) Installing libatomic1 (14.3.0-r5)
  Executing libatomic1-14.3.0-r5.post-install
(2/4) Installing libusb-1.0-0 (1.0.29-r1)
  Executing libusb-1.0-0-1.0.29-r1.post-install
(3/4) Installing libsane (1.4.0-r1)
  Executing libsane-1.4.0-r1.post-install
(4/4) Installing sane-test (1.4.0-r1)
  Executing sane-test-1.4.0-r1.post-install
OK: 11.6 MiB in 140 packages
```

tests/integration/targets/apk/tasks/main.yml
```diff
@@ -37,3 +37,27 @@
...
+            state: present
+            allow_untrusted: true
+
+        - name: "Verify package is installed"
+          community.openwrt.command:
+            cmd: apk info -e htop
+          register: apk_verify_untrusted
+          failed_when: apk_verify_untrusted.rc != 0
+
+      finally:
```
russoz: I believe you mean:
```suggestion
      always:
```
😉

tests/integration/targets/apk/tasks/main.yml
```diff
@@ -37,3 +37,27 @@
...
+        - name: Remove downloaded package
```
russoz: But instead of doing this, use the `setup_remote_tmpdir` role and it will solve this cleanup matter automatically.

russoz (commented):
Hi @Jonny007-MKD 

There's a bug - triggering the CI - but also an improvement (that happens to remove the offending line from the test ;-) )

russoz:
> yes, #195 targets both package managers. Either this PR only fixes one half of #195, or we split it into two issues.
> 
> I would not add new features for opkg, though. Until now nobody could use this feature and I would not recommend writing playbooks for opkg anymore.

You know what? You are absolutely right. I am actually going to update #195 to be only about `apk` and, if _anyone_ request support for that in `opkg`, we raise a separate issue for that.

Keep in mind that #195 was pretty much copied from `gekmihesg.openwrt` - work performed in #158.

tests/integration/targets/apk/tasks/main.yml
Jonny007-MKD: yes, I like that :+1:

Jonny007-MKD:
Oh no, the CI again! And I am too dumb to read the output. All I find is either the complete log or "Error:" :(

russoz:
Well the problem is printed in the log, somewhere. 😉

Jonny007-MKD:
Yes, but due to the nature of tests it is full with errors and failures that are expected :D

russoz:
Well, if you are talking about that pitiful state the tests were a couple of months ago, then it has been solved. All tests should be coming clean now, if they aren't then there _is_ something to be fixed. (99% of the time ;-))

russoz:
@Jonny007-MKD I have just merged some improvements in testing, running the integrations tests locally now is done with `nox` and it is quite easy. Give it a try.

```shell
nox -e test -- apk
```

Jonny007-MKD:
rebased

Jonny007-MKD:
Yes, that was really easy!
> [ERROR]: Task failed: Action failed: The module interpreter '/usr/bin/python3' was not found.

I used `ansible.builtin.command` instead of the openwrt one.

Also, I needed to fix the verify step.

russoz:
Hi @Jonny007-MKD,

You mention you need to fix the verify step. The next release is set to tomorrow evening (UTC+12), about 24h from now. I can push it further a couple of days if needed, please let me know.

Jonny007-MKD:
I ment I had to fix it, and already did with the last force push :) All tests are green ☑️

russoz:
Sorry I misunderstood it. I will take a closer look within the next couple of hours, and merge it if it's all good.

plugins/modules/apk.sh
```diff
@@ -20,12 +20,16 @@ install_packages() {
     [ -n "$pkgs_to_install" ] || return 0
 
     [ -n "$_ansible_check_mode" ] || {
-        apk $update_cache $no_cache $force_broken_world add $pkgs_to_install >"$out" 2>"$err"
+        apk $update_cache $no_cache $force_broken_world $allow_untrusted add $pkgs_to_install >>"$out" 2>"$err"
```
russoz: Why are we changing from `>"$out"` to `>>"$out"` ? From what I see, `out` is set to a tempfile (`mktemp`) and it seems to be used only to hold the immediate results of a command. 

This could lead to the module's `stdout` RV carrying more info than before, unexpectedly.

I believe this should be kept as it was before.

plugins/modules/apk.sh
```diff
@@ -20,12 +20,16 @@ install_packages() {
...
         rc=$?
         stdout="$(cat "$out")"
         stderr="$(cat "$err")"
         # Verify installation
         for pkg in $pkgs_to_install; do
+            if [ -f "$pkg" ]; then
+                pkg="$(basename "$pkg")"
+                pkg="${pkg%%-[0-9]*}"
+            fi
```
russoz, line 34: This has two problems:

- it creates an implicit rule that the filename (after basename and the %% trimming) must match exactly the name of the package that it installs. This may not be true for every cases. I had the same issue in `community.general.pipx`, the way I solved it then was to have two different parameters: `source` and `name`, effectively delegating the responsibility for figuring that thing out to the user. See https://github.com/ansible-collections/community.general/blob/main/plugins/modules/pipx.py#L51-L75 
  - you may either adopt that solution, leading to more coding, another parameter, etc... and unsatisfied users complaining about having two parameters.
  - you may keep your code as-is and add documentation explaining to the users that file and package must have equal names and the module will fail if they don't.

- packages like `ntfs-3g-low-2022.10.3-r1` (`apk fetch` gives me the file `ntfs-3g-low-2022.10.3-r1.apk`) will be parsed by this as `ntfs`, which is wrong, as it should be `ntfs-3g-low`.

russoz (commented):
Hi @Jonny007-MKD 

Still got a couple of wrinkles to iron here. I can delay the release by a couple of days, if you want to/can work on it these days. Please let me know.

plugins/modules/apk.sh
russoz: Packages that have `-[0-9]` within their name:
```shell
ssh router 'apk list' | awk '{ print $1 }' | grep -P '^[a-z]+-\d+[a-z]*-[a-z]'
```

Jonny007-MKD: Very good catch! That is indeed a problem. It doesn't affect as many packets as your grep prints, as these `-r3` suffixes belong to the version number. But a requirement that the package's filename corresponds to the package name is not desireable.

Jonny007-MKD: I am now using `apk adbdump` and parse its output using jshn. This way we can safely query the name of every apk file for verification.

Jonny007-MKD:
I found a trailing whitespace and rebased.

plugins/modules/apk.sh
russoz: > Very good catch! That is indeed a problem. It doesn't affect as many packets as your grep prints, as these `-r3` suffixes belong to the version number. But a requirement that the package's filename corresponds to the package name is not desireable.

MY regexp pride was a bit hurt, but you are right. This is a more precise one:

```shell
ssh router 'apk list' | awk '{ print $1 }' | grep -P '^[a-z]+-\d+[a-z]*-[^(r\d)]' 
```

plugins/modules/apk.sh
russoz: This one is still pending - any reason to accumulate the output here?

plugins/modules/apk.sh
russoz: > I am now using `apk adbdump` and parse its output using jshn. This way we can safely query the name of every apk file for verification.

Nice! I like that :-) thanks for updating

plugins/modules/apk.sh
russoz: Could you please return the redirection to:
```suggestion
        apk $update_cache $no_cache $force_broken_world $allow_untrusted add $pkgs_to_install >"$out" 2>"$err"
```

Jonny007-MKD: done :)

plugins/modules/apk.sh
```diff
@@ -20,12 +20,18 @@ install_packages() {
...
-        apk $update_cache $no_cache $force_broken_world add $pkgs_to_install >"$out" 2>"$err"
+        apk $update_cache $no_cache $force_broken_world $allow_untrusted add $pkgs_to_install >"$out" 2>"$err"
         rc=$?
         stdout="$(cat "$out")"
         stderr="$(cat "$err")"
         # Verify installation
         for pkg in $pkgs_to_install; do
+            if [ -f "$pkg" ]; then
+                # we installed a file and have to query the file for the package name to verify its installation
+                json_load "$(apk adbdump --format json "$pkg")" || fail "could not parse output of apk adbdump"
```
russoz, line 31: Sorry, I missed that before. I think this should be:

```suggestion
                json_load "$(apk adbdump --format json '$pkg')" || fail "could not parse output of apk adbdump"
```

Otherwise the `"` will be closing the previous quote (after `json_load`), and $pkg is effectively unquoted.

russoz: And yes, the problem kinda exists, but my suggestion does not fix it:
```
root@router:~# f() {
>   for i in $*; do
>     echo $i
>   done
> }
root@router:~# pkg="aaa bbb"
root@router:~# f "$(echo adbdump --format json '$pkg')"
adbdump
--format
json
$pkg
# it does not fix, it returns the literal `$pkg`


root@router:~# f "$(echo adbdump --format json "$pkg")"
adbdump
--format
json
aaa
bbb
# it should return one single line `aaa bbb`
```

The likelihood of the the package file ever containing a whitespace is ridiculously small, but the problem is if it contains `; rm /etc/passwd` instead of just a whitespace.

Jonny007-MKD: I think your test example is incorrect: `echo` outputs all its arguments (but removes the quotes). `"aaa bbb"` is a single argument for echo. Then the for loop in `f()` separates `aaa` from `bbb` as the output of `echo` is simply `aaa bbb` (without quotes).

`json_load` does not split the arguments, so it is safe (I believe :see_no_evil:)

Actually, `wrapper.sh` does exactly the same: https://github.com/ansible-collections/community.openwrt/blob/9db7cea314758bf9921fcfffdd6f90c569b6b0a5/plugins/modules/wrapper.sh#L17

Jonny007-MKD: This would be the correct `f()` for simulating `json_load()`:
```
f() {
  for i in "$@"; do
    echo $i
  done
}
```

russoz: And I stand corrected again - in my defence it was late at night.

As you may have noticed, I merged #221 and, with that, we now have a shorter release cadence. This change should be released in less than 2 weeks time.

plugins/modules/apk.py
```diff
@@ -55,6 +55,11 @@
       - Continue even if package dependencies are broken.
     type: bool
     default: false
+  allow_untrusted:
+    description:
+      - To install a locally available apk package, for example if this device has no internet access but you can upload apk packages directly to it.
+    type: bool
+    default: false
```
russoz, line 62: And I forgot to mention before:
```suggestion
    default: false
    version_added: 1.5.0
```

russoz (approved):
LGTM

russoz:
@Jonny007-MKD thanks for your contribution!

PR #242, opened 2026-06-11
Add uci find_all command support

vfreex:
## SUMMARY

Add support for the internal `find_all` command to `community.openwrt.uci`.

The command is already implemented by the shell backend. This change exposes it in the module metadata and returns all matching section references through `result_list`.

## ISSUE TYPE

- Feature Pull Request

## COMPONENT NAME

plugins/modules/uci.py

## ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/uci-find-all-command.yml
+++ b/changelogs/fragments/uci-find-all-command.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - uci - add ``find_all`` to the Python module metadata command choices and return its matches in ``result_list`` (https://github.com/ansible-collections/community.openwrt/pull/242).
--- a/plugins/modules/uci.py
+++ b/plugins/modules/uci.py
@@ -40,6 +40,7 @@ options:
       - del_list
       - export
       - find
+      - find_all
       - get
       - import
       - rename
@@ -166,8 +167,11 @@ result:
   type: str
   sample: cfg12523
 result_list:
-  description: The list form of result.
-  returned: when O(command=get)
+  description:
+    - The list form of C(result).
+    - For O(command=find_all), this is the list of matching positional section IDs, for example V(@wifi-iface[0]).
+      Named sections are returned as positional IDs.
+  returned: when O(command=get) or O(command=find_all)
   type: list
   sample: ["0.pool.ntp.org", "1.pool.ntp.org"]
 config:
--- a/tests/integration/targets/uci/tasks/main.yml
+++ b/tests/integration/targets/uci/tasks/main.yml
@@ -59,6 +59,21 @@
     option: testopt
     value: testval
 
+- name: "find_all - setup: add another matching test section"
+  community.openwrt.uci:
+    command: add
+    config: uhttpd
+    type: testtype
+  register: uci_add_second_section
+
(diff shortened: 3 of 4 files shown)


changelogs/fragments/uci-find-all-command.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - uci - add ``find_all`` to the Python module metadata command choices and return its result as a list.
```
russoz: Please add the URL for this PR by the end of the description.

-------
Please refer to the standard formatting for the changelog fragments' entries:

https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to-format

The outcome of what you write will be published in the collection changelog:

https://docs.ansible.com/ansible/latest/collections/community/general/changelog.html

plugins/modules/uci.py
```diff
@@ -161,9 +162,11 @@
...
 result:
-  description: Output of the C(uci) command.
+  description:
+    - Output of the C(uci) command.
+    - For O(command=find_all), this is a list of positional section IDs, for example V(@wifi-iface[0]).
   returned: always
-  type: str
+  type: raw
   sample: cfg12523
 result_list:
```
russoz: Please use `result_list` to carry the result of that operation instead of `result`.

changelogs/fragments/uci-find-all-command.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
russoz: Technically this is a new feature, since the docs never mentioned that option, even though it is implemented in the code. I checked in the gekmihesg.openwrt repo and that mistake was brought as-is from over there :-D

So please change this to `minor_changes:` and change the description of the PR to "Feature Pull Request".

tests/integration/targets/uci/tasks/main.yml
```diff
@@ -76,6 +91,80 @@
...
+
+- name: "find_all - matches found: assert result contains all matching sections"
+  ansible.builtin.assert:
+    that:
+      - uci_find_all_match is succeeded
+      - uci_find_all_match.result is defined
+      - uci_find_all_match.result is not string
+      - uci_find_all_match.result | length >= 2
+      - uci_add_section.result in uci_find_all_match.result
+      - uci_add_second_section.result in uci_find_all_match.result
```
russoz: Please include an assertion for the values as well. Here it is only proving that something _was_ found. Let's ensure that it found the right thing.

russoz (commented):
Hi @vfreex 

Thank you for your contribution!! It needs a couple of adjustments, but overall is looking good!

Also, please update your main branch and rebase this branch with the latest from main (to include #244) - otherwise tests will fail for the combination Ansible devel + Python 3.12.

Thanks!

vfreex:
@russoz Thanks for the review! I've addressed your review suggestions and updated the PR. PTAL.

russoz:
Hi @vfreex

Sorry for taking long to review, I have had a couple of busy weeks. I am planning on taking a closer look until the end of the week.

russoz (approved):
LGTM

PR #243, opened 2026-06-12
feat: remove legacy param parsing

### LINKED ISSUE #234: Remove legacy parsing of parameters
### Summary

As discovered in #233, the Ansible machinery will actually scan the module file for the literal text `WANT_JSON`. If found, then the machinery will provide the parameters in JSON format, otherwise it will be legacy.

Given that 
- `wrapper.sh` always had `WANT_JSON` in its content, thus ensuring the params to be JSON, that remains true after #233 is merged
- we have no specific need for the legacy format

There is no point in retaining code that parses parameters passed in the legacy style.

Thus, this issue calls for the removal of that code. It should wait for #233 to be merged first.

### Issue Type

Feature Idea

### Component Name

wrapper.sh

### Additional Information

```yaml (paste below)

```

### Code of Conduct

russoz:
## SUMMARY

Remove legacy param parsing as it is not really used.

Fixes #234 

## ISSUE TYPE

- Feature Pull Request

## COMPONENT NAME

plugins/module_utils/_core.sh

## ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/_core.sh
+++ b/plugins/module_utils/_core.sh
@@ -121,36 +121,7 @@ _verify_value_type() {
     echo "$_value"
 }
 
-_parse_legacy_params() {
-    local _var _type _required _default _alias
-    local _param _value _IFS
-    for _param in $PARAMS; do
-        eval "${_param%%[/=]*}=\"\""
-    done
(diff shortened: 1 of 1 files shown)


russoz:
Rebase after #244

PR #248, opened 2026-07-10
opkg: add `force_feeds` parameter

### LINKED ISSUE #239: opkg: add force_feeds parameter for vendor-feed override
## Summary

Add a `force_feeds` (or equivalent) parameter to `community.openwrt.opkg` so the module can override the default feed discovery when the host's `opkg` is configured to use vendor-specific feeds that don't include the packages being installed.

## Component

opkg

### Problem

Teltonika's RUTX/RUT/TRB firmware ships a modified `opkg` binary that defaults to Teltonika-only feeds, even when `/etc/opkg/distfeeds.conf` is a symlink pointing at an OpenWrt feed file. Installing OpenWrt packages (e.g. the `python3-*` packages from `packages-21.02`) on these devices therefore requires explicitly passing `-e <distfeeds.conf>` to `opkg` at the command line — symlinking the distfeeds file is necessary but not sufficient.

The `community.openwrt.opkg` module has no parameter to pass this flag (or its long form, `--force_feeds`). The module's `DOCUMENTATION` block lists only `name`, `state`, `force`, `update_cache`, `autoremove`, `nodeps`. The only options today are:

1. Use `community.openwrt.command` with the `-e` flag in the cmd string (loses the module's state-aware idempotency, requires manual `creates:` / `changed_when:` handling).
2. Edit `/etc/opkg.conf` itself to point at the OpenWrt feeds, which on Teltonika firmware may be reverted by the firmware or by the vendor's opkg wrapper.

### Proposed solution

Add a `force_feeds` parameter to `community.openwrt.opkg` that, when set, passes `-e <path>` (or `--force_feeds <path>`) to the underlying opkg invocation. Mirroring the long form from `opkg`'s CLI would match Teltonika's own error message wording ("If you want to use OpenWrt feeds, run opkg with `--force_feeds /etc/opkg/openwrt/distfeeds.conf`"), so I'd suggest:

```yaml
- name: Install OpenWrt 21.02 packages on a Teltonika device
  community.openwrt.opkg:
    name: python3-base,python3-light,python3-ctypes,python3-openssl
    state: present
    force_feeds: /etc/opkg/openwrt/distfeeds.conf
```

Backwards compatibility: when `force_feeds` is unset, behavior should be unchanged from the current module.

### Reproducer / motivation

A role that provisions Teltonika RUTX/RUT/TRB devices hit this exact issue on 2026-06-03 when refactoring its bootstrap install tasks from `community.openwrt.command` to `community.openwrt.opkg`. The previous `command` invocation was:

```yaml
- name: Router | Bootstrap Python 3.9 packages (can take a while)
  community.openwrt.command:
    cmd: >-
      opkg -e /etc/opkg/openwrt/distfeeds.conf install
      python3-base python3-light python3-codecs python3-ctypes
      python3-logging python3-multiprocessing python3-email python3-urllib
      python3-distutils python3-decimal python3-xml
      python3-openssl
```

Replacing it with `community.openwrt.opkg: name: …, state: present` (no `force_feeds` parameter) caused the next run to fail with:

```
opkg install  python3-logging: Unknown package ' python3-logging'.
The default package feeds only include software supported by Teltonika.
If you want to use OpenWrt feeds, run opkg with
'--force_feeds /etc/opkg/openwrt/distfeeds.conf'.
```

A `force_feeds` parameter would have let us keep the module's clean state semantics and idempotency while still passing the vendor-required override.

### Versions

- `community.openwrt` collection: verified by reading `~/.ansible/collections/ansible_collections/community/openwrt/plugins/modules/opkg.py` on `main` and confirming no `force_feeds` / `distfeeds` / `opkg_conf` / `config_file` parameter exists in `DOCUMENTATION`.
- Tested on Teltonika RUTX11 firmware `07.14.x` (RUTX_R_00.07.14.x) running Teltonika's stock `opkg` build.
- OpenWrt 21.02 is the upstream target for the role's `python3-*` package set.

### Code of Conduct

I agree to follow the [Ansible Code of Conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).

russoz:
## SUMMARY

Add a `force_feeds` parameter to `community.openwrt.opkg` that passes `--force_feeds <path>` to `opkg install` and `opkg update` invocations.

This allows users to override the default feed configuration — specifically needed on Teltonika firmware, where the stock `opkg` binary defaults to Teltonika-only feeds. Removing a package does not consult feeds, so the flag is intentionally not applied to `opkg remove`.

Fixes #239

## ISSUE TYPE

- Feature Pull Request

## COMPONENT NAME

opkg

## ADDITIONAL INFORMATION

Tested pattern: when `force_feeds` is unset, behaviour is identical to before (no flag is added).

### DIFF
--- a/changelogs/fragments/248-opkg-force-feeds.yml
+++ b/changelogs/fragments/248-opkg-force-feeds.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "opkg - add ``force_feeds`` parameter to pass ``--force_feeds`` to opkg install and update invocations (https://github.com/ansible-collections/community.openwrt/issues/239, https://github.com/ansible-collections/community.openwrt/pull/248)."
--- a/changelogs/fragments/248-opkg-force-feeds.yml
+++ b/changelogs/fragments/248-opkg-force-feeds.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - "opkg - add ``force_feeds`` parameter to pass ``--force_feeds`` to opkg install and update invocations (https://github.com/ansible-collections/community.openwrt/issues/239, https://github.com/ansible-collections/community.openwrt/pull/248)."
+  - "opkg - add ``conf_file`` parameter to pass ``--conf`` to opkg install and update invocations (https://github.com/ansible-collections/community.openwrt/issues/239, https://github.com/ansible-collections/community.openwrt/pull/248)."
--- a/plugins/modules/opkg.py
+++ b/plugins/modules/opkg.py
@@ -65,6 +65,12 @@ options:
     description:
       - Do not follow dependencies.
     type: bool
+  force_feeds:
+    description:
+      - Path to a feeds configuration file, passed to opkg as C(--force_feeds).
+      - Useful on devices where the default opkg uses vendor-only feeds (for example, Teltonika firmware).
+    type: str
+    version_added: 1.7.0
 """
 
 EXAMPLES = r"""
@@ -94,6 +100,12 @@ EXAMPLES = r"""
     name: busybox
     state: present
     force: reinstall
+
+- name: Install a package using OpenWrt feeds on a Teltonika device
+  community.openwrt.opkg:
+    name: python3-base
+    state: present
+    force_feeds: /etc/opkg/openwrt/distfeeds.conf
 """
 
 RETURN = r""""""
--- a/plugins/modules/opkg.py
+++ b/plugins/modules/opkg.py
@@ -65,10 +65,13 @@ options:
     description:
       - Do not follow dependencies.
     type: bool
-  force_feeds:
+  conf_file:
     description:
-      - Path to a feeds configuration file, passed to opkg as C(--force_feeds).
-      - Useful on devices where the default opkg uses vendor-only feeds (for example, Teltonika firmware).
+      - Path to a custom opkg configuration file, passed to opkg as C(--conf).
+      - This replaces the default opkg configuration entirely instead of adding to it, so the file must define
+        its own C(dest) and C(lists_dir) entries in addition to any C(src) feed lines it needs.
+      - Useful on devices where the default opkg configuration does not provide the desired feeds (for example,
+        Teltonika firmware).
     type: str
     version_added: 1.7.0
 """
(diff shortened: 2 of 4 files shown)


Jonny007-MKD (approved):

russoz:
CI was failing on every Integration job because `--force_feeds <path>` is not a real opkg option — it is a Teltonika-specific extension to their forked opkg build (see the CLI error quoted in #239), not part of upstream/stock opkg. Since the Integration jobs run against `ghcr.io/openwrt/rootfs` images (stock opkg), every job rejected the flag outright with `opkg: unrecognized option: force_feeds`.

The parameter has been renamed `force_feeds` → `conf_file`, and it now passes opkg's own standard `--conf <conf_file>` option instead. This achieves the same practical goal — pointing opkg at a different feeds configuration — using an option that exists in stock opkg (and is very likely retained by Teltonika's fork alongside their own `-e`/`--force_feeds` shortcut).

One behavioral caveat worth calling out: unlike the default lookup (which merges `/etc/opkg.conf` with every `*.conf` file under `/etc/opkg/`), `--conf` replaces the opkg configuration entirely rather than adding to it. So the file passed to `conf_file` must be self-contained — it needs its own `dest` and `lists_dir` entries in addition to any `src` feed lines. Docs were updated to spell this out.

The integration test was rewritten to actually exercise that behavior instead of just re-pointing at the same default `distfeeds.conf` (which isn't a complete config on its own, and was never truly exercised — the old test always died earlier on the unrecognized flag). It now builds a self-contained custom config on the fly, reusing the device's own feed URLs but aimed at a fresh, empty lists directory, and checks that this produces a genuinely different, verifiable outcome: install fails before an `opkg update` against that config, succeeds afterward, and stays idempotent on a third run.

russoz:
Hi @Jonny007-MKD the tests were failing when you approved the PR :-) I have twisted the purpose of the PR a little bit, but now there is a solution that works for any system. And the test are passing ;-)

Jonny007-MKD (approved):

Jonny007-MKD:
Thank you for your work :)

russoz:
@Jonny007-MKD thanks for reviewing!

PR #263, opened 2026-09-11
uci: only set a scalar option when the value differs

### LINKED ISSUE #262: uci: command=set always reports changed, even when the option already holds the target value
### Summary

`community.openwrt.uci` with `command=set` reports `changed` on every run for scalar options, even when the option already holds the target value. Convergence is correct, and diff mode correctly shows an empty diff, but the `changed` flag is wrong.

The practical cost is handlers. A role that sets scalar options and notifies a reload handler reloads the service on every run, because every task reports changed.

### Issue Type

Bug Report

### Component Name

uci

### Ansible Version

```
ansible [core 2.19.5]
```

### Community.OpenWrt Version

```
community.openwrt 1.7.0
```

### OpenWrt Version

```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='21.02.0'
DISTRIB_ARCH='mipsel_24kc'
```

Tested on a Teltonika RUT951, firmware RUT9M_R_00.07.24.3. The root cause is in `libuci`, so I expect the same result on supported OpenWrt releases.

### Steps to Reproduce

```yaml
- hosts: routers
  gather_facts: false
  roles:
    - community.openwrt.init
  tasks:
    - name: Read the current hostname
      community.openwrt.uci:
        command: get
        key: system.@system[0].hostname
      register: current

    - name: Set the hostname to the value it already has
      community.openwrt.uci:
        command: set
        key: system.@system[0].hostname
        value: "{{ current.result }}"
```

### Expected Results

The second task reports `ok`, because the option already holds the target value.

### Actual Results

The second task reports `changed`, and the diff is empty.

```
TASK [Read the current hostname] ***********************************************
ok: [router]

TASK [Set the hostname to the value it already has] ****************************
changed: [router]
```

### Root cause

`uci_set` sets scalars unconditionally:

```sh
uci_set() {
    ...
        *) try "uci set \"\$key=\$$var\"";;
    esac
}
```

`cleanup()` then decides `changed` by comparing an md5 of `uci changes` before and after:

```sh
[ "$changes" = "$(uci_change_hash)" ] || {
    changed
    ...
}
```

`uci set` records a delta even when the new value equals the current value. This is `libuci` behaviour, not a module bug. I confirmed it on the device with an isolated delta directory, so the real pending state stayed untouched:

```
# uci changes
# uci -P /tmp/probe set system.@system[0].hostname=RUT951
# uci -P /tmp/probe changes
system.system.hostname='RUT951'
```

The delta is recorded, `uci_change_hash` moves, and the module reports changed.

`uci_set_list` does not have this problem. It calls `uci_compare_list` first and returns early on a match. Only the scalar branch of `uci_set`, and the scalar branch of `uci_set_dict`, set without comparing.

This affects `command=set`, and also `command=ensure` and `command=section`, because both call `uci_ensure`, which calls `uci_set`.

### Proposed fix

Compare before setting, which matches what `uci_set_list` already does. I checked the logic on the device:

```
value already matches  -> delta: []
value differs          -> delta: [system.system.hostname='NEWNAME']
```

I am happy to raise a pull request for this.

gjkraken:
## SUMMARY

`command=set` reported `changed` on every run, even when the option already held the target
value. `uci_set` writes unconditionally, and `cleanup()` derives `changed` from an md5 of
`uci changes`. `libuci` records a delta even when the new value matches the current one, so
the flag was always set. This also affected `command=ensure` and `command=section`, since
both reach `uci_set` through `uci_ensure`.

This adds `uci_compare_value`, called before the write, mirroring the early return that
`uci_compare_list` already does for list options.

Assisted-by: Claude Opus 5

Fixes #262

## ISSUE TYPE
- Bugfix Pull Request

## COMPONENT NAME

uci

## ADDITIONAL INFORMATION

Comparing with `uci get` is not safe, because it joins list values with a space:

```paste below
# uci show uhttpd.probe.opt
uhttpd.probe.opt='a' 'b'
# uci get uhttpd.probe.opt
a b
```

Setting that option to the scalar `a b` would compare equal, so the module would skip the
write and leave a list in place. `uci_compare_value` therefore uses `uci_get_safe`, which
returns the quoted form, and requires exactly one word after expansion.

Every individual write is of a scalar option, including the entries of a dict passed to
`command=section`, because `uci_set_dict` sets them one at a time. So dict values are
covered too.

Review follow-ups:

- I took the intent of the quoting comment but not the exact form. The single-quoted version
  builds the same string, which I checked, but `shellcheck` flags it as SC2016 and that code
  has no entry in the ignore files. Instead `uci_set` now resolves the parameter into
  `set_value` and uses plain expansions, which removes the escaped quotes and matches how
  `uci_set_dict` already calls `try uci set`.
- Added a task that sets `max_connections` to a different value before the first assertion,
  so it no longer depends on the container image default.
- Documented the change in the module `notes`, alongside the other behaviour notes in this collection.
- Rebased onto `main`. The `Sanity (Ⓐdevel)` failure was the SC2319 errors in `_core.sh`,
  `package_facts.sh` and `setup.sh`, fixed by #266.

Ran locally against containers:

```paste below
$ nox -e test -- uci
openwrt2305  : ok=41  changed=18  unreachable=0  failed=0  ignored=1
openwrt2410  : ok=41  changed=18  unreachable=0  failed=0  ignored=1
openwrt2512  : ok=41  changed=18  unreachable=0  failed=0  ignored=1

$ andebox tox-test -e dev -- sanity --docker default
congratulations :)

$ nox -e lint
Ran 5 sessions: formatters, codeqa, yamllint, antsibull-nox-config, lint - all success
```


tests/integration/targets/uci/tasks/main.yml
```diff
@@ -44,6 +44,136 @@
...
       - uci_delete is succeeded
 
+- name: "idempotency - setup: set an option to a known value"
+  community.openwrt.uci:
+    command: set
+    config: uhttpd
+    section: main
+    option: max_connections
+    value: "150"
+  register: uci_idem_first
```
russoz, line 62: For the sake of completeness, please set that option to a different value _before_ this task, so that the test ensures that setting to 150 does cause change.

tests/integration/targets/uci/tasks/main.yml
```diff
@@ -44,6 +44,136 @@
...
+- name: "idempotency - set the same section dict again"
+  community.openwrt.uci:
+    command: section
+    config: uhttpd
+    type: testidem
+    name: idemsection
+    value:
+      opta: aaa
+      optb: bbb
+  register: uci_idem_dict_repeat
```
russoz, line 109: Technically a dict is not a scalar, so your changelog fragment needs adapting. Also, the PR description mentions that as well.

changelogs/fragments/262-uci-idempotent-set.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - uci - compare the current value before setting a scalar option, so that ``command=set``, ``command=ensure`` and ``command=section``
+    no longer report ``changed`` when the option already holds the target value
```
russoz:
```suggestion
  - uci - compare the current value before setting a scalar option, so that ``command=set``, ``command=ensure`` and ``command=section`` are idempotent
```

plugins/modules/uci.sh
```diff
@@ -135,7 +146,8 @@ uci_set() {
             json_select_real "$var"
             uci_set_dict
             json_select ..;;
-        *) try "uci set \"\$key=\$$var\"";;
+        *) eval "uci_compare_value \"\$key\" \"\$$var\"" ||
+                try "uci set \"\$key=\$$var\"";;
```
russoz: Wouldn't it be simpler and more readable like:
```suggestion
        *) eval 'uci_compare_value "$key" "$'"$var"\" ||
                try 'uci set "$key=$'"$var"\";;
```
?

TBH I have not tested - I just try to avoid `"... \"..."` as much as I can.

russoz (commented):
Hi @gjkraken,

Thanks for your contribution!!

It looks good overall, just a couple of comments (in code, plus here):
- You can run the tests locally using `nox`. The easiest way is to use the devcontainer configured for the project, and run `nox -e test -- uci`. If you prefer not to use devcontainers, you can set up a local venv for you, installing `requirements-test.txt` and running that same command.
- It kinda looks like this PR was generated by an AI. No problems with that, but please note that, when using AI, you must abide by the [Ansible Community Policy for AI-Assisted Contributions](https://docs.ansible.com/projects/ansible/latest/community/ai_policy.html#ansible-community-policy-for-ai-assisted-contributions). Essentially your commits/PR should declare using the AI, which model, etc.
- Also, whether AI or not, please do use the PR template for this repository. We do have some tools that use that pre-defined format to extract information. When the description looks completely different from that, they fail.
- I have released the tests (as you are a first-time contributor here, I have to manually unblock them), and please notice that there are some sanity tests failing. You can run them locally with `nox` as well.

Feel free to disagree with any of the comments, we can always reason about them.

Once again, thank you!
Cheers

russoz:
The PR should also add something to the docs to let the users know that these uci commands are idempotent since version 1.8.0

gjkraken:
Hey @russoz thanks for the review and guidance, I've followed up on all your comments, let me know if there's anything outstanding and I can follow up: 

- Local tests: Set up a venv from `requirements-test.txt` and ran everything in containers:
`nox -e test -- uci` passes on 23.05.6, 24.10.8 and 25.12.4 (`ok=41 changed=18 failed=0`), `andebox tox-test -e dev -- sanity --docker default` is clean, and `nox -e lint` passes.
 
- Sanity failures: Those were not from this PR. The four errors were SC2319 in `_core.sh`, `package_facts.sh` and `setup.sh`, none of which this PR touches, and `main` was red on the scheduled runs until #266 landed. I rebased onto `main` and sanity is green.

- Quoting: Your version builds a byte-identical string, which I verified, but `shellcheck` then reports `uci.sh:150:21: SC2016: Expressions don't expand in single quotes`, and SC2016 has no entry in the ignore files. Rather than add one, `uci_set` now resolves the parameter into `set_value` first and uses plain expansions. That removes the escaped quotes and hopefully helps with readability, and has the side benefit of making the scalar branch call `try uci set` the same way `uci_set_dict` does. Happy to go a different way if you prefer.
 
- Dict is not a scalar: You already fixed the fragment, thanks. I reworded the PR description so it no longer frames this as scalar-only, and noted that dict entries are covered because `uci_set_dict` sets them one at a time.
 
- Test completeness: Added a task that sets `max_connections` to a different value before the first assertion, so it no longer depends on the image default.
 
- Docs: Added a note to the module description that `set`, `ensure` and `section` are idempotent as of 1.8.0.
 
- AI policy: Fair call, and my apologies for not declaring it up front. The commits now carry `Assisted-by: Claude Opus 5`, and the PR description says the same. I have reviewed and tested everything here myself.

Also rewrote the description using the repository template.

Cheers!

changelogs/fragments/262-uci-idempotent-set.yml
```diff
@@ -0,0 +1,3 @@
+bugfixes:
+  - uci - compare the current value before setting a scalar option, so that ``command=set``, ``command=ensure`` and ``command=section`` are idempotent
```
russoz:
```suggestion
  - uci - compare the current value before setting a scalar or dict option, so that ``command=set``, ``command=ensure`` and ``command=section`` are idempotent
```

russoz:
1Hi @gjkraken 

Thanks for your continued work here.

>     * Sanity failures: Those were not from this PR. The four errors were SC2319 in `_core.sh`, `package_facts.sh` and `setup.sh`, none of which this PR touches, and `main` was red on the scheduled runs until [multiple: fix shellcheck `SC2319` warnings on `$?` capture #266](https://github.com/ansible-collections/community.openwrt/pull/266) landed. I rebased onto `main` and sanity is green.

Yeah, I could not reproduce them locally, but the CI was failing everytime. In any case, now the scripts are no longer offending that check.

>     * Quoting: Your version builds a byte-identical string, which I verified, but `shellcheck` then reports `uci.sh:150:21: SC2016: Expressions don't expand in single quotes`, and SC2016 has no entry in the ignore files. Rather than add one, `uci_set` now resolves the parameter into `set_value` first and uses plain expansions. That removes the escaped quotes and hopefully helps with readability, and has the side benefit of making the scalar branch call `try uci set` the same way `uci_set_dict` does. Happy to go a different way if you prefer.

Nah, your solution is much better than my suggestion. Thanks!

>     * Dict is not a scalar: You already fixed the fragment, thanks. I reworded the PR description so it no longer frames this as scalar-only, and noted that dict entries are covered because `uci_set_dict` sets them one at a time.

Actually my suggestion to the changelog fragment was something else. I just dropped another one to address this issue.

>     * AI policy: Fair call, and my apologies for not declaring it up front. The commits now carry `Assisted-by: Claude Opus 5`, and the PR description says the same. I have reviewed and tested everything here myself.

No worries, I believe this should be better advertised and I keep postponing that. One day, one day.

> Also rewrote the description using the repository template.

Thanks!

Other than that last touch on the changelog fragment, it LGTM.

russoz:
Galaxy is behaving again. Re-running the failed job.

russoz:
@gjkraken thanks for your contribution!!

PR #264, opened 2026-09-12
feat(uc_uci): ucode-based UCI module with shared helper library

### LINKED ISSUE #261: Ucode as an implementation language
### Summary

I had have many times troubles with uci plugin caused by long diff (it produces incorrect/truncated json).
So thinking, why to bother with json manipulation inside shell, when you universally always have ucode + ucode-mod-uci / ubus on any modern build (i.e. with fw4).

Secondly, i'd prefer to have an option to hide sensitive fileds from uci diffs. E.g. wireguard private keys from the `network` config.

### Issue Type

Feature Idea

### Component Name

uci

### Additional Information

_No response_

### Code of Conduct

vooon:
## SUMMARY

Add `uc_uci`, an alternative to the shell uci module implemented in ucode,
emitting Ansible module JSON directly to avoid the JSON corruption that
affects long diffs in the shell/jshn path.

Introduce a shared ucode helper library (ansible_common.uc) and a reusable
action base (UcodeOpenwrtActionBase) that ships the helper next to the module
and runs it via 'ucode <module> <args>', with no shell wrapper involved.

uc_uci supports the full uci command set plus:
- redact_keys to mask sensitive options in diffs
- operations: run a list of operations in a single invocation
- displayable list diffs with UCI-export-style headers (config.section=type)
  rendered by the shared ac_upsert_section/ac_push_diff helpers

Includes integration and unit tests, a ucode linter (uc-lint.mjs), CI/sanity
wiring, documentation, and ucode conventions in AI/AGENTS.md.

Fixes #261

## ISSUE TYPE

## COMPONENT NAME

AI/AGENTS.md
antsibull-nox.toml
plugins/action/uc_uci.py
plugins/module_utils/ansible_common.uc
plugins/modules/uc_uci.uc
plugins/modules/uc_uci.yml


russoz:
@vooon TBH I had heard of ucode before but always postponed diving into it. Now that I see this PR, I see the potential here is much bigger than I ever imagined. I think this is likely a first step towards moving the entire collection to `ucode` and deprecate `sh` modules altogether - but that will require a careful analysis.

Thanks for stepping forward with this!

AI/AGENTS.md
```diff
@@ -50,3 +50,53 @@ The collection will support Ansible 2.18 onwards.
...
+`.py`. It runs as a `non_native_want_json` script: Ansible passes the args as a JSON file whose
+path is `ARGV[0]`, and the module prints a JSON result on `stdout`.
+
+Modules share a single helper library at `plugins/module_utils/ansible_common.uc` (functions
+prefixed `ac_`), and a reusable action base at `plugins/plugin_utils/ucode_action.py`
+(`UcodeOpenwrtActionBase`). The action plugin transfers the module and the helper into the same
+remote directory so the module's relative `import { ... } from './ansible_common.uc'` resolves,
+then executes `ucode <module> <args>` and parses the JSON result. No shell wrapper is involved.
+
+Hard-learned ucode rules:
```
russoz: Token waste:
```suggestion
ucode rules:
```
The AI will not value your effort ;-)

AI/AGENTS.md
russoz, line 1: I should be placing AGENTS.md in the root directory, as that seems to be more standard. That being said, the existing content is ridiculous - I wrote it - and there's probably very little that should be saved from it, if anything. I did write a AGENTS.md more recently for community.general - that one is much better and I will probably use as basis for a new one here.

Just giving a heads up that things are likely going to change in this vicinity.

AI/AGENTS.md
```diff
@@ -50,3 +50,53 @@ The collection will support Ansible 2.18 onwards.
...
+
+- `'use strict';` at the top; `#!/usr/bin/ucode` shebang and a `WANT_JSON` marker.
```
russoz: The `use strict` should be replaced with passing `-s` when invoking `ucode`, and ensuring that ALL code is strict. If some module later on needs it to be not strict for a good reason, then we may start talking about making that optional. I would prefer to start with enforced from day one.

AI/AGENTS.md
```diff
@@ -50,3 +50,53 @@ The collection will support Ansible 2.18 onwards.
 # Definition of Success
 
 - The collection must pass the molecule tests
+
+# Ucode modules
```
russoz: I would argue that this topic should be warranted its own .md file, maybe `AI/ucode.md`? WDYT?

AI/AGENTS.md
```diff
@@ -50,3 +50,53 @@ The collection will support Ansible 2.18 onwards.
...
+
+Beyond the shell-based modules, the collection ships modules written in OpenWrt's native
+`ucode` language (see `docs/docsite/rst/mod_dev_guide.rst`). A ucode module is a `.uc` file in
+`plugins/modules/` plus a `.yml` sidecar with `DOCUMENTATION`/`EXAMPLES`/`RETURN`; there is no
+`.py`. It runs as a `non_native_want_json` script: Ansible passes the args as a JSON file whose
+path is `ARGV[0]`, and the module prints a JSON result on `stdout`.
+
+Modules share a single helper library at `plugins/module_utils/ansible_common.uc` (functions
+prefixed `ac_`), and a reusable action base at `plugins/plugin_utils/ucode_action.py`
```
russoz: I am not sure that we need that prefix, but will not fight it either. The import mechanism in ucode supports aliasing so I suppose any conflict could be handled with that:
```ucode
import { default as defVal } from "./files/test3.uc";
```

AI/AGENTS.md
```diff
@@ -50,3 +50,53 @@ The collection will support Ansible 2.18 onwards.
...
+(`UcodeOpenwrtActionBase`). The action plugin transfers the module and the helper into the same
+remote directory so the module's relative `import { ... } from './ansible_common.uc'` resolves,
+then executes `ucode <module> <args>` and parses the JSON result. No shell wrapper is involved.
```
russoz: Thin line here - I know we have to tell the AI what things are not/not to do, but in my experience is better to increase the positive reinforcement than the negative one.
```suggestion
then executes `ucode <module> <args>` and parses the JSON result.
```

AI/AGENTS.md
```diff
@@ -50,3 +50,53 @@ The collection will support Ansible 2.18 onwards.
...
+- Arrays use global helpers: `length(arr)`, `push(arr, ...)`, `sort(arr)`. There are no
+  `arr.push()`/`arr.sort()` methods.
+- Strings are not `[]`-indexable — use `substr(s, i, 1)` / `ord(s, i)`.
+- `for (let x in arr)` yields the elements; over objects it yields the keys. Loop variables must
+  be declared with `let`. Iterate objects with `for (let k in obj)`; iterate arrays by index.
+- No `String(x)` global — use `sprintf("%s", x)`.
+- `uci`: `import { cursor } from 'uci'; const u = cursor();` — `get/get_all/set/foreach/add/
+  save/commit`. `set(config, section, type)` creates a named section; `add` requires the config
+  to be explicitly `load()`ed first; `save()` persists changes as a delta so a subsequent module
+  invocation (new process) sees them.
```
russoz: That sounds like standard ucode, maybe we should point the AI to ingest https://ucode.mein.io/ for the syntax/standard library stuff? From what I have been looking, that site is not complete in terms of language syntax and constructs, but it is a good start.

If all else fails, we might write up new documentation on ucode and point the AI to consume it. In any case, I think that would be better than cramming everything ucode into the AGENTS file. WDYT?

AI/AGENTS.md
```diff
@@ -50,3 +50,53 @@ The collection will support Ansible 2.18 onwards.
...
+- Check mode: the args include `_ansible_check_mode`; compute the diff/changed but skip
+  `save`/`commit`.
+- Idempotency: strip UCI meta keys (`.name`, `.type`, `.anonymous`, `.index`) from `get_all`
+  output before comparing; compare via sorted-key equality (`sprintf("%J", ...)`), not raw object
+  equality.
+
+Lint ucode modules with `node tests/uc-lint.mjs` (or the `ucode-lint` nox session /
+pre-commit hook). Run unit tests with `ansible-test units`; run a module's integration target with
+`nox -e test -- <target>`.
```
russoz: This information on tests is covered in:
docs/docsite/rst/testing_guide.rst

Better to point the AI to that one single place instead of repeating the same information in two different places.

changelogs/fragments/261-uc_uci.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - uc_uci - new ucode-based module to control OpenWrt UCI, an alternative to the shell ``uci`` module that emits Ansible module JSON directly (avoids JSON corruption on long diffs). Supports the same commands plus ``redact_keys`` to mask sensitive options in diffs and ``operations`` to run multiple operations in a single invocation (https://github.com/ansible-collections/community.openwrt/issues/261).
```
russoz: We should add the URL of the PR as well.
```suggestion
  - uc_uci - new ucode-based module to control OpenWrt UCI, an alternative to the shell ``uci`` module that emits Ansible module JSON directly (avoids JSON corruption on long diffs). Supports the same commands plus ``redact_keys`` to mask sensitive options in diffs and ``operations`` to run multiple operations in a single invocation (https://github.com/ansible-collections/community.openwrt/issues/261, https://github.com/ansible-collections/community.openwrt/pull/264).
```

docs/docsite/rst/mod_dev_guide.rst
```diff
@@ -413,3 +413,118 @@ Otherwise, the module will automatically bail out if executed in check mode.
 
 
 .. versionadded:: 0.3.0
+
+Ucode modules
+^^^^^^^^^^^^^
```
russoz, line 404: That version added refers to this document. It sho0uld remain as the last element of the file.

docs/docsite/rst/mod_dev_guide.rst
```diff
@@ -413,3 +413,118 @@ Otherwise, the module will automatically bail out if executed in check mode.
...
+
+Check mode and diff support
+"""""""""""""""""""""""""""
+
+Ucode modules honour check mode (skip persisting changes when ``_ansible_check_mode`` is set)
+and diff mode (produce a structured ``diff`` when ``_ansible_diff`` is set and the task has
+``diff: true``). The ``ac_result``/``ac_exit`` helpers carry the ``changed``/``failed``/``msg``
+contract automatically.
+
+.. versionadded:: 1.9.0
```
russoz: And this line must go away - also the next version to be released in 1.8.0

noxfile.py
russoz, line 1: I am not sure about this one:
- There is no lint for other file types in the `noxfile`, if we add this, we should add the others for consistency
- That implies installing nodejs - locally and/or in the devcontainer setup. That's slightly bigger than just adding the task in `nox`
- The entry in `pre-commit` already provides that capacity, in a more contained way.

I would defer this for later, in a separate PR.

plugins/module_utils/_ansible_common.uc
russoz, line 1: This file should have a prefix `_` to indicate that it is not to be reused by other collections straight from here. If that need arises some day, then we can revisit this issue.

The point being: if this is a "public" module, in theory it may be used by code outside this repository and any future change to this code's API will require deprecation, with a period until removal, etc... Much simpler to keep this "private" to this collection.

Also, I am not sure of the name "ansible_common" here - this is code supporting Ansible modules, I find it a bit redundant to name it like that. But I don't have a suggestion right now.

plugins/module_utils/ansible_common.uc
```diff
@@ -0,0 +1,286 @@
...
+// modules. Modules `import { ... } from './ansible_common.uc'` and use these
+// helpers to implement the Ansible module contract (result object, failure,
+// change detection), plus ucode/UCI utilities.
+//
+// ucode notes (see docs/docsite/rst/mod_dev_guide.rst):
+//   - `export function name(...) {...};` must end with a semicolon.
+//   - Functions are not hoisted: define before use, no forward declarations.
+//   - Arrays use the global helpers (push, sort, length); not array methods.
+//   - Strings are not []-indexable; use substr(s, i, 1).
+//   - JSON: decode with json(str), encode with sprintf("%J", obj).
```
russoz: Repeating these things here is redundant

plugins/module_utils/ansible_common.uc
```diff
@@ -0,0 +1,286 @@
+// Copyright (c) Ansible Project
+// GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+// ansible_common.uc — shared helper library for ucode-based community.openwrt
+// modules. Modules `import { ... } from './ansible_common.uc'` and use these
+// helpers to implement the Ansible module contract (result object, failure,
+// change detection), plus ucode/UCI utilities.
```
russoz: The ucode code in this comment brings no value. 
```suggestion
// ansible_common.uc — shared helper library for ucode-based modules. 
// Modules use these helpers to implement the Ansible module contract 
// (result object, failure, change detection), plus ucode/UCI utilities.
```

plugins/module_utils/ansible_common.uc
```diff
@@ -0,0 +1,286 @@
...
+
+'use strict';
+
+import { readfile } from 'fs';
+
+// ---- argument / value coercion -------------------------------------------
+
+// Coerce a JSON value to a ucode boolean. Accepts bool, int and common
+// string representations ("true", "yes", "1", "on").
+export function ac_bool(v) {
```
russoz: I would ditch the comment and rename the function:
```suggestion
export function coerce_to_bool(v) {
```
(maybe keep the `ac_` prefix, or not, that'd be a different conversation)

plugins/module_utils/_ansible_common.uc
```diff
@@ -0,0 +1,286 @@
...
+	let ak = sort(keys(a));
+	let bk = sort(keys(b));
+	if (length(ak) != length(bk))
+		return false;
+	for (let i = 0; i < length(ak); i++) {
+		if (ak[i] != bk[i])
+			return false;
+		if (!ac_same(a[ak[i]], b[bk[i]]))
+			return false;
+	}
```
russoz, line 241: I would test the lists directly: if `b` is equal to `a` except for one missing entry and one extra entry, the length is still going to be the same.

Then thge loop could be simplified to:
```suggestion
	if (ak != bk)
		return false;
	for (key, value in a){
		if (!ac_same(a[key], b[key]))
			return false;
	}
```

plugins/modules/uc_uci.uc
```diff
@@ -0,0 +1,581 @@
+#!/usr/bin/ucode
+// WANT_JSON — args passed as a JSON file whose path is ARGV[0]; output on stdout.
+// Copyright (c) 2026, Alexei Znamensky (@russoz)
```
russoz: "It wasn't me" -- Homer Simpson

russoz (commented):
Hi @vooon 

Thanks for your contribution!

A first round of comments.

AI/AGENTS.md
russoz: This is happening in #265

russoz:
Beware the Ansible Galaxy is misbehaving for the last hour or so. Sometime it is just a matter of re-running the failing jobs.

Update: I just triggered the re-run

AI/AGENTS.md
vooon: Well, i just copied section from my other ucode project, where model had to re-learn ucode from source many times, else it tries to write it as on JS. :)

AI/AGENTS.md
vooon: Probbaly, but my agents usually not any better...

AI/AGENTS.md
vooon: Maybe, but models do not learn "-s" flag, when they sees 'use strict'. Also that's worth for node-based linter.

AI/AGENTS.md
vooon: I prefer to place into root AGENTS simply because models i'm using tend to forget that ucode is not JS :)

AI/AGENTS.md
vooon: That's actually good question what style better to use.
If to use aliasing for each imported function - the import became too long.
The same time it's possible just to `import * as ac from "./ansible_common.uc";`
And at the same time, good-old-prefix helps with possible stdlib shadowing.

AI/AGENTS.md
vooon: Oh, that's from my first experiments when it tries to use the same approach as shell uses, and failed...

AI/AGENTS.md
vooon: Well, in other projects i've tried just to point to docs, but found "laziness", until you don't just ask RTFM :)
So having small set of "don't" and "not javascript" lines saved a lot of re-reads. But yes, we may shorten the list.
There also one big gotcha: the site corresponds to the latest ucode, which you can find in current openwrt master, 25.12.x lacking some of them - e.g. forward declarations!

noxfile.py
vooon: That's actually quick and dirty linter just to catch problems llms often does.
Just allows faster reconciliation loop. I'd wish to have better tooling...
Btw github ci always have node available because of actions impl language.

plugins/module_utils/_ansible_common.uc
vooon: We may do that, but actually i disagree to say that "module_utils" as a public module.
I always threat them as a private part of the collection (look at usual python import pathes).
And if you extract some part of it into your project - it's your trouble to maintain.

plugins/module_utils/_ansible_common.uc
vooon: It mostly based on https://ucode.mein.io/tutorial-05-dictionaries.html#dictionary-equality 
Ideally i'd make unit tests to cover all utils, but that requires ucode on the CI. Previously i've done that by custom action.

plugins/modules/uc_uci.uc
vooon: The funny thing with that generated code that i hardly can say that i'm wrote it too :)
That feels more like you architect and just pass all the lifting to your fellow mid to complete :)

vooon:
Found much better linter/lsp: https://github.com/NoahBPeterson/ucode-lsp

vooon:
I'm not sure, but what if to minify sources transferred on the fly? I.e. embed common and strip all comments?

vooon:
@russoz i've addressed most of your comments, but not all. And there a lots of changes.
Now it introduce the library closer to the ansible python's one, in exchange to library size.
On-the-fly minification still might be useful, but actually in my observations more of the time ansible spends on scp handshakes, than actually transfer (at least on on 10k ~ 100kiB).

vooon:
Oh, wow. I should have made that plugin while ago, `wrt_firewall` role now lightning fast :)

AI/AGENTS.md
russoz: Of course, in AGENTS we point to it, justlikewe point to docs

noxfile.py
russoz: Yes, but not everyone wants to install nodejs (and the 54836464657 dependencies for every package in their computers.

plugins/module_utils/_ansible_common.uc
russoz: It is the convention in Ansible collections. Thus comes from Python, where `_names` are "private" and should not be imported (outside the collection). 

The message we send with this here is that: this module should not be imported by code outside the collection.

vooon:
One interesting finding is about module output truncation - i thought that maybe jshn cannot handle large jsons, but actually the real fix is the `\n` placed before json report!
I've added regression test for that scenario. Possibly shell wrapper needs similar trick.

AI/AGENTS.md
russoz: FYI, I have just merged #265

russoz: And I correct myself, the flag is `-S` (uppercase) for strict.

AI/AGENTS.md
russoz: There's not a huge ecosystem of modules and functions in the ucode universe. I don't think that name conflict is really going to be an issue.

AI/AGENTS.md
russoz: That's a good catch! We should definitely point it out to the LLM.

As of laziness... well, I know what you mean, but in my experience that is fruit of the model used. The free and lazy models will always be... cheap and lazy. I am not sure that putting the bar that low is the best policy (naturally, making it a super high bar for it also doesn't help). I have no final conclusion on this, I suppose this is one of those situations where we go with whatever we got, and as time goes by we adjust as needed.

plugins/module_utils/_ansible_common.uc
russoz: You can create a new molecule scenario at the collection level, under `extensions/molecule`. The CI workflow might need adjustments to accommodate the new test.

OTOH, this PR is already not small - I would recommend deferring that to a next PR.

plugins/modules/uc_uci.uc
russoz: That's _exactly_ how those things are working these days.

russoz:
> I'm not sure, but what if to minify sources transferred on the fly? I.e. embed common and strip all comments?

To what purpose? Reduce the size? That could be an optional feature, for a future PR.

plugins/module_utils/_ansible_common.uc
russoz: > It mostly based on https://ucode.mein.io/tutorial-05-dictionaries.html#dictionary-equality 

I had seen that example before, but given the facts that `keys(a)`/`keys(b)` _must_ both be lists of strings (no variation on that), testing the (sorted) lists directly should give us the results of both tests in L232-237. With one check.

vooon:
@russoz 
> To what purpose? Reduce the size?

Mainly I thought to have as minimal amount of file transfers as possible (i.e. amount ssh handshakes) .
But actually, after the porting my roles from ansible `loop:` to `operations: [...]`, speedup is a such high, so I do not mind about few extra kilobytes on JSDoc comments.

I've made several reviews by Sol 5.6 and Opus 5 and fixed most of caveats. Added more correctness tests.
Hopefully that would not require yet another force pushes...

AI/AGENTS.md
vooon: Switched to `* as ac`.

russoz:
Force pushes are not required. All commits are squashed upon merging. 

(yeah, this case was an exception because it was a rebase with conflict)

AGENTS.md
```diff
@@ -39,6 +39,18 @@
   `docs/docsite/rst/mod_dev_guide.rst`. Read it before touching modules, action plugins, or
   `wrapper.sh`, rather than relying on assumptions from other Ansible collections.
 
+### Ucode modules
+
+The collection also ships ucode modules (`plugins/modules/*.uc` + a `.yml` sidecar), which run
```
russoz: This is standard Ansible, no need to explain it. 
```suggestion
The collection also ships ucode modules (`plugins/modules/*.uc`), which run
```

AGENTS.md
```diff
@@ -39,6 +39,18 @@
...
+with OpenWrt's native `ucode` interpreter. Before writing one, read
+`docs/docsite/rst/ucode.rst` — it links the official ucode docs and records the rules that apply
+on the supported OpenWrt releases (note the 25.12.x version caveats, e.g. no forward
```
russoz:
```suggestion
on the supported OpenWrt releases (note the 25.12.x version caveats, for example no forward
```

Please refer to the [linguistic guidelines](https://docs.ansible.com/projects/ansible/latest/dev_guide/style_guide/index.html#linguistic-guidelines) for the documentation, as well as the guide to [preferred spelling](https://docs.ansible.com/projects/ansible/latest/dev_guide/style_guide/spelling_word_choice.html).

docs/docsite/rst/ucode.rst
```diff
@@ -0,0 +1,184 @@
...
+  used; there is no ``export function name;`` forward declaration. The reference documentation
+  (master) shows examples that rely on hoisting that will **not** run on 25.12.
+- When in doubt, test on the oldest supported target (currently ``25.12``) rather than assuming
+  the reference examples work.
+
+Collection conventions
+----------------------
+
+A ucode module is a ``.uc`` file in ``plugins/modules/`` plus a ``.yml`` sidecar with
+``DOCUMENTATION``/``EXAMPLES``/``RETURN``; there is no ``.py``. Modules run as
```
russoz:
```suggestion
``DOCUMENTATION``/``EXAMPLES``/``RETURN``; there is no ``.py``. Modules run as
```
Ansible allows docs to be placed in a.py file, in fact that's what the shell modules are doing. Come to think of it, I think it would make more sense to have them using yaml YAML files, but it is not for this specific document to establish that generic standard.

vooon:
I just prefer making better git history. I have also split this work in a separate commits for better reviewability.

docs/docsite/rst/ucode.rst
```diff
@@ -0,0 +1,184 @@
...
+``non_native_want_json`` scripts: Ansible passes the args as a JSON file whose path is
+``ARGV[0]``, and the module prints a JSON result on ``stdout``.
```
russoz, line 43: That was a bit over:
- Ansible modules can be documented in YAML files or Python files. 
- Explaining the internals of Ansible in the ucode documentation (`non_native_want_json`?? I never heard or worried about that before, and the shell modules probably use the same mechanism).
- Users and module developers should not need to worry about how the input/output of the module (unless they bump in to issues and need to troubleshoot, of course). When writing modules in Python there is no need to worry about JSON. Or even in the shell modules, most (but not all) of them hide that completely from the developer.

```suggestion
A ucode module is a ``.uc`` file in ``plugins/modules/``.
```

docs/docsite/rst/ucode.rst
```diff
@@ -0,0 +1,184 @@
...
+
+The reusable action base at ``plugins/plugin_utils/ucode_action.py``
+(``UcodeOpenwrtActionBase``) transfers the module and its helper library into the same remote
+directory so the module's relative import resolves, then executes ``ucode <module> <args>`` and
+parses the JSON result. No shell wrapper is involved.
+
```
russoz, line 47: Over-detailed.
```suggestion
The reusable action base at ``plugins/plugin_utils/ucode_action.py``
(``UcodeOpenwrtActionBase``) MUST be used by all ucode modules, otherwise Ansible is unable to execute `ucode`.

```

docs/docsite/rst/ucode.rst
```diff
@@ -0,0 +1,184 @@
...
+namespace. It exports helpers for the Ansible module contract, argument coercion, change
+detection and diff redaction:
+
+.. code-block:: text
+
+  ac.load_args()
+      Read and parse the module args from ``ARGV[0]`` (the JSON args file).
+
+  ac.new_result()
+      Build the standard ``{changed, failed, msg}`` result object.
```
russoz, line 92: Maintaining code documentation away from the code is not a good idea. Please remove this from here.

docs/docsite/rst/ucode.rst
```diff
@@ -0,0 +1,184 @@
...
+container: ``tests/unit/ucode/`` (``test_*.uc`` scripts staged with ``_ansible_common.uc``) run
+via the ``ucode_unit`` nox session or ``tests/unit/ucode/run.sh``.
+
+Check mode and diff support
+---------------------------
+
+Ucode modules honour check mode (skip persisting changes when ``_ansible_check_mode`` is set)
+and diff mode (produce a structured ``diff`` when ``_ansible_diff`` is set and the task has
+``diff: true``). The ``ac.new_result``/``ac.exit_json`` helpers carry the
+``changed``/``failed``/``msg`` contract automatically.
```
russoz, line 182: This is specific to the uci module. We may - in fact I plan to - use ucode for running other commands. Each module manages its own check/diff modes. Please remove this.

plugins/module_utils/_ansible_common.uc
```diff
@@ -0,0 +1,631 @@
...
+		before = strip_meta(cur);
+
+	// Normalise a wanted scalar to the string form UCI persists: booleans as
+	// 0/1, ints as their decimal string. Lists are kept as-is (UCI list
+	// options). This keeps the compare and the write consistent with UCI's
+	// string-typed storage, so an int/bool input no longer re-diffs every run.
+	let uci_value = function(v) {
+		let t = type(v);
+		if (t == 'bool')
+			return v ? '1' : '0';
```
russoz, line 370: Simpler as
```suggestion
			return int(v);
```

plugins/module_utils/_ansible_common.uc
```diff
@@ -0,0 +1,631 @@
...
+ * via redact().
+ * @param {object} u uci cursor
+ * @param {string} config config name
+ * @param {string} sid section id (or section name)
+ * @param {string} sec_type section type
+ * @param {object} want desired option map
+ * @param {UpsertOptions} [opts] behaviour flags (see UpsertOptions)
+ * @returns {{before: object, after: object, changed: boolean}}
+ */
+export function upsert_section(u, config, sid, sec_type, want, opts) {
```
russoz, line 355: Please rename all these functions that operate with the uci cursor to `uci_xxxxxxx`.

plugins/module_utils/_ansible_common.uc
```diff
@@ -0,0 +1,631 @@
...
+};
+
+// ---- Ansible module result contract --------------------------------------
+
+/**
+ * Render a ucode template string (Jinja-style ``{{ ... }}`` / ``{% ... %}``)
+ * against the given scope and return the output as a string. Uses ucode's
+ * native template engine via ``loadstring(str, { raw_mode: false })`` +
+ * ``render()``; template variables are read from the global scope, so the
+ * scope dict is bound globally for the duration of the render and restored
```
russoz, line 290: Where is that global scope coming from? The code refers to `global` (object), where is that one defined?

Using global variables is usually a poor practice, any compelling reason why it should be used here?

In fact, this `global` variable is not used or mentioned anywhere else in this PR. I strongly recommend this feature to be removed.

plugins/module_utils/_ansible_common.uc
```diff
@@ -0,0 +1,631 @@
...
+ * afterwards. Returns an empty string on error.
+ * @param {string|null} str template source
+ * @param {object} scope template variables
+ * @returns {string}
+ */
+export function render_template(str, scope) {
+	if (str == null)
+		return '';
+	if (type(str) != 'string')
+		return '';
```
russoz, line 300: If the developer passes something that is not a string to render a string template, that should pop an error instead of silently return `""`. And by doing that, the `null` test above is rendered redundant as well.

russoz:
@vooon I am sending a batch of comments on this, but I would recommend you to hold on a bit. As I mentioned before, your initiative on this triggered something, and I am working on a rewrite for this collection. We should have an `AnsibleModule` object, just like the Python modules (actually not exactly like them, but close), plus other perks.

I would very much like you to contribute the `uci` module using the new code, but it is still a little bit early.

russoz (commented):
As mentioned, some comments I have written down before.

docs/docsite/rst/ucode.rst
vooon: Yes, that's the same as `.sh` works, but i actually spent some time trying to understand if i can make pipelining to work because `ansible.builtin.copy` much-much faster than `community.openwrt.copy` (noticed because now i've dropped python off the firmwares...).

Note that `.ps1` plugins supports pipelining, but it's unlikely to push `.uc` pipelining patch to the ansible-core...

AGENTS.md
vooon: Oh, nice docs. Don't think i ever read them :)

plugins/module_utils/_ansible_common.uc
vooon: No, we should convert bools into strings, else we would have wrong changed=true (and diff). UCI stores only strings.

vooon:
Replaced by #280

PR #268, opened 2026-09-13
feat(wg_plugins): add wg_interface and wg_peer ucode modules

### LINKED ISSUE #261: Ucode as an implementation language

vooon:
## SUMMARY

Add two ucode-based modules for managing WireGuard/AmneziaWG:

- `wg_interface` manages identity keys (via ubus wireguard/amneziawg
      genkey/pubkey/genpsk), the interface UCI section (including AWG 3.x/3.1
      options when proto is amneziawg) and the WAN ingress firewall rule for the
      interface listen port.
      Interfaces share a single host identity key while each peer gets a distinct pre-shared key,
      simplifying monitoring mapping.
      Interface/peer section ids and the firewall rule name/id are rendered from
      ucode templates.

- `wg_peer` writes WireGuard/AmneziaWG peer UCI sections for an interface,
      using key material supplied by `wg_interface`.

Both modules build on the shared ac_* helper library: ac_render_template for
    templated ids, ac_upsert_section for idempotent section writes, ac_push_diff
    for list diffs, and ac_result(extra) for pre-initialized result fields.

Includes integration tests (wireguard + amneziawg p2p), sanity ignores,
    changelog fragment and updated module developer docs.

Fixes #261 

## ISSUE TYPE

## COMPONENT NAME

AI/AGENTS.md
antsibull-nox.toml
plugins/action/uc_uci.py
plugins/action/wg_interface.py
plugins/action/wg_peer.py
plugins/module_utils/ansible_common.uc
plugins/modules/uc_uci.uc
plugins/modules/uc_uci.yml
plugins/modules/wg_interface.uc
plugins/modules/wg_interface.yml
plugins/modules/wg_peer.uc
plugins/modules/wg_peer.yml

## ADDITIONAL INFORMATION
This two plugins feeds my needs to setup many p2p links for my OSPF VPN Mesh and basically may be done using `command` and `uci` plugins.
But it saves a lot of SSH transfers by batching: instead of many plugin transfers you do only few calls:
1. `wg_interface` - to create/update interface and resolve host private key, psks for each peer.
2. `wg_peer` - to create/update all peers
3. possible handler runs for interface restart / firewall restart.

Please note, that it's intentionally made to keep the same private key on all the managed interfaces - that simplifies a lot Grafana dashboard, when all of your 10's p2p links have same public key.
While still every peer have it's own PSK key, so you won't be able to have wrong peer to connect to the someone's else's device.
And secondly - module made that way, that host private_key never leaves the host, and never printed in diff.
That solves accidential key exposure to LLMs :)

And at the same time those modules generic enough for other scenarios, like managing more generic multi-peer hub-and-spoke setups.


vooon:
@russoz i want to test "gh stack", but seems it only works on my fork, or i just doing something wrong :)
My local molecule runs real VMs with ping checks, but unlikely i could get beyond config checks here, with docker.
I also unsure, if it's needed to anyone else. But that speedups my vpn playbook drastically. That was actually my driving reason to explore ucode plugins :)

```
Checking existing stacks...
Looking up PRs for 2 branches...
⚠ failed to update base branch for PR #268 to uc-uci-with-batching: HTTP 422: Validation Failed
PullRequest.base is invalid
⚠ Stacked PRs are not enabled for this repository
zsh: exit 9     gh stack link  
```

russoz:
I have never uaed gh stack before. The duplications from the other PR should not be in this one, but we'll tend to that after the ucode PR is merged. 

And I will have follow-up ucode PRs after that one as well.

vooon:
Superseded by #281, which ports the WireGuard modules onto the ucode/UCI foundation from #280.

PR #272, opened 2026-09-21
multiple: add a framework for modules written in ucode, and convert apk, command, opkg and ping

russoz:
## SUMMARY

This PR enables modules to be written in [ucode](https://github.com/jow-/ucode), OpenWrt's own scripting
language, which ships in the base image of every supported release. A ucode module is a `.uc` file
with its documentation in a `.yml` sidecar; its action plugin transfers the module together with the
module utils it declares, and runs it through `ucode_wrapper`.

`basic.uc` provides the module side of the contract, modelled on `AnsibleModule`: argument parsing
and validation (type, elements, required, default, choices, aliases, and the checks that span
several parameters), check and diff mode, a result object, `run_command()`, and
`exit_json()`/`fail_json()`. The validation follows ansible-core, including the wording of its
messages and the order it applies them in, so that a parameter which merely declares a default does
not count as given for `mutually_exclusive`.

`apk`, `command`, `opkg` and `ping` are converted to ucode, carrying their logic over unchanged, and
their documentation moves to sidecars. `wait_for_connection` drives `ping`, so its action plugin
follows, and its documentation moves as well.

Relates to #261

## ISSUE TYPE
- Feature Pull Request
- Refactoring Pull Request

## COMPONENT NAME

antsibull-nox.toml
plugins/action/apk.py
plugins/action/command.py
plugins/action/opkg.py
plugins/action/ping.py
plugins/action/wait_for_connection.py
plugins/module_utils/basic.uc
plugins/modules/apk.py
plugins/modules/apk.sh
plugins/modules/apk.uc
plugins/modules/apk.yml
plugins/modules/command.py
plugins/modules/command.sh
plugins/modules/command.uc
plugins/modules/command.yml
plugins/modules/opkg.py
plugins/modules/opkg.sh
plugins/modules/opkg.uc
plugins/modules/opkg.yml
plugins/modules/ping.py
plugins/modules/ping.sh
plugins/modules/ping.uc
plugins/modules/ping.yml
plugins/modules/ucode_wrapper.sh
plugins/modules/ucode_wrapper.yml
plugins/modules/wait_for_connection.py
plugins/modules/wait_for_connection.yml

## ADDITIONAL INFORMATION

Tested with the sanity suite, and with the integration tests for every module.

The modules were also exercised directly against the `ghcr.io/openwrt/rootfs` images for the three
supported releases, which carry a real `ucode`, `opkg` and `apk`.


.github/BOTMETA.yml
```diff
@@ -21,13 +21,13 @@ files:
   $modules/:
     ignored: gekmihesg
     maintainers: $team_repo_mgrs
-  $modules/apk.py:
+  $modules/apk.uc:
     maintainers: kbialek
```
russoz:
```suggestion
    maintainers: kbialek russoz
```

.github/BOTMETA.yml
```diff
@@ -21,13 +21,13 @@ files:
...
-  $modules/apk.sh:
+  $modules/apk.yml:
     maintainers: kbialek
```
russoz:
```suggestion
    maintainers: kbialek russoz
```

russoz:
Hi @vooon please take a look at this PR whenever you can.

plugins/modules/apk.uc
```diff
@@ -0,0 +1,143 @@
...
+
+// The packages to act upon, given as a comma-separated list.
+function requested_packages() {
+    let pkgs = [];
+    for (let pkg in split(params.name, ',')) {
+        if (pkg != '')
+            push(pkgs, pkg);
+    }
+    return pkgs;
+}
```
Jonny007-MKD: isn't this the same as coercing as a list (maybe with removing empty elements)?

plugins/modules/apk.uc
```diff
@@ -0,0 +1,143 @@
...
+
+    let dump = module.run_command([ 'apk', 'adbdump', '--format', 'json', pkg ]);
+    let meta;
+    try {
+        meta = json(dump.stdout);
+    } catch (e) {
+        meta = null;
+    }
+    if (type(meta) != 'object' || type(meta.info) != 'object' || meta.info.name == null)
+        module.fail_json('could not parse output of apk adbdump');
```
Jonny007-MKD: maybe add information about the file that failed?

plugins/modules/command.uc
```diff
@@ -0,0 +1,86 @@
...
+    },
+});
+
+const params = module.params;
+const result = module.result;
+
+// Quote a string for use as a single shell word.
+function shell_quote(value) {
+    return `'${replace(sprintf('%s', value), /'/g, "'\\''")}'`;
+}
```
Jonny007-MKD: this function already exists in basic.uc

plugins/modules/command.uc
```diff
@@ -0,0 +1,86 @@
...
+
+// ---- main -----------------------------------------------------------------
+
+result.update({ cmd: params.cmd, start: '', end: '', delta: '', stdout: '', stderr: '', rc: 0 });
+
+if (params.chdir != null && !chdir(params.chdir))
+    module.fail_json(`cd ${params.chdir}: unable to change directory`);
+
+if (params.creates != null && exists(params.creates))
+    module.exit_json({ stdout: `skipped, since ${params.creates} exists` });
```
Jonny007-MKD: is there no special skipped-flag that we have to set?

plugins/modules/command.uc
```diff
@@ -0,0 +1,86 @@
...
+    res = module.run_command(`echo ${shell_quote(params.cmd)} | xargs sh -c 'exec "$@"' --`);
+else
+    res = module.run_command([ params.executable, '-c', params.cmd ]);
+
+let ts_end = time();
+
+result.update({
+    start: timestamp(ts_start),
+    end: timestamp(ts_end),
+    delta: elapsed(ts_end - ts_start),
```
Jonny007-MKD: `time()` returns the current epoch in seconds, which (afaik) may be affected by NTP updates etc. Using a monotonic clock for durations is best practice, and ucode provides `clock(true)`. As system may not provide a monotonic clock, though, so a fallback should be implemented as well.

Maybe this was an issue with the shell implementation as well, but with ucode it's rather easy to improve it. I'm fine if we do this later, though :)

Jonny007-MKD (commented):
ucode is definitely more readable than shell scripts :) Is there any other advantage?

russoz:
> ucode is definitely more readable than shell scripts :) Is there any other advantage?

from the top of my head (I'm sure the is more):

- handling lists and dicts is natural, instead of convoluted and error-prone
- given that ^ it is much easier now to provide a deprecation mechanism
- it is faster
- it has bindings directly into `uci`, rather than juggling CLI calls
- it has allowed to - more easily - implement proper param validations

;-)

plugins/modules/command.uc
russoz: This was picked up on the development, it is git stashed for later ;-) There will be many PRs for bugfixes and improvements later on.

plugins/modules/ucode_wrapper.sh
vooon: IMHO wrapper is not needed, especially if the only job - to set `-S` and `-L`.
And it's actually makes behaviour much worse on high-latency links (i.e. 250ms ping - each file transfer counts, while the size is negligible).

vooon: And ucode module directly can have WANT_JSON marker.

plugins/module_utils/basic.uc
```diff
@@ -0,0 +1,615 @@
...
+// ---- result output --------------------------------------------------------
+
+// Print a result object as JSON and terminate the module, exiting non-zero
+// when the result says the module failed. A result without a `failed` status
+// is a bug in the caller, and dies loudly rather than assuming success.
+function emit(result) {
+    if (result.failed == null)
+        die('result is missing the "failed" status');
+
+    printf('%J\n', result);
```
vooon:
```suggestion
    printf('\n%J\n', result);
```

Leading `\n` is essential to fix long responses (>8 kiB).
Just try to make diff of ~128-256 kiB of data (see my regression test).

plugins/module_utils/basic.uc
```diff
@@ -0,0 +1,615 @@
...
+                collection_name: COLLECTION_NAME,
+            });
+        },
+
+        // Snapshot the accumulated fields for output, with `extra` merged on top.
+        render: function(extra) {
+            return extra ? { ...data, ...extra } : { ...data };
+        },
+    };
+}
```
vooon: I'd add diff helper.

plugins/modules/command.uc
```diff
@@ -0,0 +1,86 @@
...
+const params = module.params;
+const result = module.result;
+
+// Quote a string for use as a single shell word.
+function shell_quote(value) {
+    return `'${replace(sprintf('%s', value), /'/g, "'\\''")}'`;
+}
+
+// Whether a filename or glob pattern matches anything.
+function exists(pattern) {
```
vooon: Func name may conflict with core [`exists()`](https://ucode.mein.io/module-core.html#exists).

vooon (commented):
AnsibleModule() is nice abstraction.

plugins/modules/ucode_wrapper.sh
vooon: AFAIC it also supports a shebang, if you want to stay using `_execute_module()`.

russoz: We have to, and I quote:

This is the ActionBase API for an action plugin to invoke an Ansible module:

```python
result = self._execute_module(
    module_name="ansible.builtin.command",
    module_args={"cmd": "something"},
    task_vars=task_vars,
)
```

It does considerably more than merely "run the module". It handles things such as:
- module resolution/loading
- templating/module arguments
- creating the remote temporary directory
- transferring the module / AnsiballZ payload
- permissions
- pipelining
- become
- connection handling
- async wrapping
- executing the module
- collecting and decoding its result


----
That being said, -S in the shebang is easy, but the -L might give us some trouble.

plugins/modules/command.uc
russoz: :facepalm: fixed

plugins/modules/apk.uc
russoz: Pretty much - but it is mimicking the existing behaviour of the shell module. I believe it makes sense to change the parameter to a `list` later on, but I would prefer to add any improvements in later PRs. This one is already big as it is.

plugins/modules/apk.uc
russoz: I believe that is the exact same behaviour we had before (with the shell module) - I will check things later today (my TZ). If it replicates the behaviour, I intend to keep as-is, then later on we move in with improvements.

plugins/modules/command.uc
russoz: Not sure what you mean by that. This is kinda standard behaviour of `command`, even in `ansible.builtin`.

plugins/module_utils/basic.uc
russoz: I am trying to hold back on functional improvements in this PR - changing the implementation language is already a big thing, I do not want to make it bigger.

That will be a nice addition to it, I agree.

plugins/modules/apk.uc
russoz: On second thought, we can mimic the exact same behaviour with a more concise idiom. Just pushed the change.

plugins/modules/ucode_wrapper.sh
russoz: > IMHO wrapper is not needed, especially if the only job - to set `-S` and `-L`. And it's actually makes behaviour much worse on high-latency links (i.e. 250ms ping - each file transfer counts, while the size is negligible).

I have been investigating this a bit. 

I would rather get the ucode support off the ground as-is - no worse than the shell version of it - and then we can start working on optimising the module execution later on.

plugins/module_utils/basic.uc
russoz: I would also like this issue to be worked on after the initial support is merged. I have just released version 1.8.0 yesterday, so if we merge this PR in the next couple of days we have almost two weeks to sort things out, and when 1.9.0 gets released, it will hopefully carry all these improvements.

WDYT?

russoz: On second thought, this one goes in now rather than later. Some digging backs the suggestion up:

- There is precedent right here: the shell framework has always emitted a blank line before its JSON — `plugins/module_utils/_core.sh:92` is `echo; json_dump`. The ucode path quietly dropped a workaround the shell path considered necessary.
- The mechanism fits: ansible-core parses module output with `_filter_non_json_lines()` (`lib/ansible/module_utils/json_utils.py`), which scans for the first line *starting* with `{` and gives up with "No start of json char found" otherwise. Anything that ends up sharing that line — interleaved stderr, transport noise — breaks the parse, and the size threshold is consistent with output crossing a buffer boundary where interleaving becomes possible. The leading newline guarantees the brace opens a line; the filter drops leading blank lines by construction, so it costs nothing.

Unlike the other deferred items, this is a latent correctness bug rather than a missing feature, and the fix is one character — so it is in, thanks for catching it.

russoz: That was my buddy Claudy shaming me in public. :stuck_out_tongue_closed_eyes: 

Later on, I will want to include that test here as well. Just for kicks.

plugins/modules/ucode_wrapper.sh
vooon: Ok, this is just a bits of my pain points, why i start thinking on optimizations.

plugins/module_utils/basic.uc
vooon: Ok.

plugins/modules/ucode_wrapper.sh
russoz: Happy to address those, I am just trying not to cram everything into one single PR.

vooon (approved):
I do not see any new things, i'd need to comment.
Looks good to me!

plugins/modules/command.uc
Jonny007-MKD: ah, ok :+1: I thought we'd have to signal that the action was skipped, just like with changed or success

Jonny007-MKD (approved):
same for me :)

russoz:
I made some minor changes after the approvals - tests passed. Will assume things are still OK, and we have more than 10 days yet to the day of the next release, so there's plenty of time to fix any lingering snafu.

Thanks @vooon and @Jonny007-MKD for reviewing!!

Jonny007-MKD:
I'm not happy with the string-to-float workaround. Is this really the best that ucode can do? :(

russoz:
@Jonny007-MKD yeah, it is not a very clean take. 

One thing we are missing is unit testing for ucode code. I got some ideas but once we have that, we can more confidently list all the testcases and try different implementations.

